PR #257, opened 2015-03-03
Fix #235: /configuration resource doesn't support basic authentication

### LINKED ISSUE #235: /configuration resource doesn't support basic authentication
Is it expected ?

schaubl:
Refactored the RootApplication to make sure all resources are using the same method to attach authenticators.
Corresponding issue #235.

### DIFF
--- a/jar-service/src/main/java/com/sixsq/slipstream/application/RootApplication.java
+++ b/jar-service/src/main/java/com/sixsq/slipstream/application/RootApplication.java
@@ -140,6 +140,7 @@ public class RootApplication extends Application {
 
 	private void logServerStarted() {
 		String message = "Server started";
+		Logger.debug(message);
 		Logger.info(message);
 		Logger.warning(message);
 		Logger.severe(message);
@@ -255,7 +256,8 @@ public class RootApplication extends Application {
 	}
 
 	private void attachConfiguration(RootRouter router) {
-		guardAndAttach(router, ServiceConfigurationResource.class, ServiceConfigurationResource.CONFIGURATION_PATH);
+		String rootUri = ServiceConfigurationResource.CONFIGURATION_PATH.replaceAll("^/",	"");
+		guardAndAttach(router, ServiceConfigurationResource.class, rootUri);
 	}
 
 	private void attachLogout(RootRouter router) {
@@ -347,10 +349,10 @@ public class RootApplication extends Application {
 	}
 
 	private void attachDocumentation(RootRouter router) {
-		Authenticators authenticators = guardAndAttach(router, DocumentationResource.class, "/documentation")
+		Authenticators authenticators = guardAndAttach(router, DocumentationResource.class, "documentation")
 				.getAuthenticators();
 
-		for (ListIterator<Authenticator> iter = authenticators.listIterator(); iter.hasNext();) {
+		for (ListIterator<Authenticator> iter = authenticators.listIterator(); iter.hasNext(); ) {
 			iter.next().setOptional(true);
 		}
 	}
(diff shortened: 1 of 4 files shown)


jar-service/src/main/java/com/sixsq/slipstream/application/RootApplication.java
```diff
@@ -351,39 +349,24 @@ private void attachAction(RootRouter router) {
 	}
```
konstan: Shouldn't attachAction() follow the introduced pattern and use guardAndAttach() as well?

jar-service/src/main/java/com/sixsq/slipstream/application/RootApplication.java
```diff
@@ -231,19 +251,13 @@ private Directory attachDownloadsDirectory() {
 		return directory;
 	}
 
-	private void attachReports(RootRouter router)
-			throws ConfigurationException, ValidationException {
+	private void attachReports(RootRouter router) throws ConfigurationException, ValidationException {
 		router.attach("/reports", new ReportRouter(getContext(), router.getApplication()));
 	}
```
konstan: Shouldn't attachReports() follow the introduced pattern and use guardAndAttach() as well?

konstan:
Thanks for the clarification.  Just checked on the last provided link.  There are two comments.

jar-service/src/main/java/com/sixsq/slipstream/application/RootApplication.java
schaubl: I don't know what is the ActionResource but it seems to be for `Reset Password` and `User Email Validation`.
See https://github.com/slipstream/SlipStreamServer/blob/issue_%23235/jar-service/src/main/java/com/sixsq/slipstream/action/OneShotActionPerformer.java#L41
So this resource should not have any authentication.

jar-service/src/main/java/com/sixsq/slipstream/application/RootApplication.java
schaubl: attachReports() is a bit special, it does his authentication inside the ReportRouter.
This is why I didn't touch it.
I will try to see if we can use guardAndAttach() for this.

konstan: Is `/` (forward slash) OK in this case in the root URI?

jar-service/src/main/java/com/sixsq/slipstream/application/RootApplication.java
konstan: Thanks. 
Is `/` (forward slash) OK in this case in the root URI?

jar-service/src/main/java/com/sixsq/slipstream/application/RootApplication.java
schaubl: I checked the code of `ReportRouter` and this code is not similar to all others resources because it adds a `Authorizer` and a `Filter`.
So I don't know if&how we can safely extract (all | part of) the content of the `ReportRouter` to `attachReports()`.

jar-service/src/main/java/com/sixsq/slipstream/application/RootApplication.java
schaubl: Yes, because we are using it directly into `router.attach(...)` instead of first passing it to `convertToRouterRoot(...)`.

jar-service/src/main/java/com/sixsq/slipstream/application/RootApplication.java
schaubl: > Is / (forward slash) OK in this case in the root URI?
Yes, because we are using it directly into router.attach(...) instead of first passing it to convertToRouterRoot(...).

PR #263, opened 2015-03-10
Fix #262 - Metrics are send even if the user has no Cloud credentials

### LINKED ISSUE #262: Metrics are sent even if the user has no Cloud credentials
Because empty metrics are sent, graphite (carbon-cache) creates a file which uses space on the disk.

schaubl:
Metrics are send even if the user has no Cloud credentials.
Because empty metrics are sent, graphite (carbon-cache) creates a file which uses space on the disk.

@st : Can you review this ?
Maybe `(updator/update user)` can be called directly instead of adding it into the channel from this channel. What do you think ?

Corresponding issue: #262


schaubl:
The last commit above was made with @st regarding to his suggestions.

PR #268, opened 2015-03-12
Fix #250 - Allow empty ssh key when creating run with an extra parameter

### LINKED ISSUE #250: Allow empty ssh key when creating run
If an ssh key is required for the cloud, create a temporary one.

Linked to: slipstream/SlipStreamUI#182

schaubl:
Corresponding issue: #250.

### DIFF
--- a/jar-connector/src/main/java/com/sixsq/slipstream/connector/ConnectorBase.java
+++ b/jar-connector/src/main/java/com/sixsq/slipstream/connector/ConnectorBase.java
@@ -294,18 +294,31 @@ public abstract class ConnectorBase implements Connector {
         return privateSshKeyFile;
     }
 
+    public static String getServerPublicSshKeyFilename() throws ConfigurationException, ValidationException {
+		return Configuration.getInstance().getProperty(ServiceConfiguration.CLOUD_CONNECTOR_ORCHESTRATOR_PUBLICSSHKEY);
+    }
+
     private String getUserPublicSshKey(User user) throws ValidationException, IOException {
 		return user.getParameter(
 				ExecutionControlUserParametersFactory.CATEGORY + "." + UserParametersFactoryBase.SSHKEY_PARAMETER_NAME)
 				.getValue();
     }
 
+    private String getUserOrServerPublicSshKey(User user) throws ValidationException, IOException {
+		String userPublicSshKey = getUserPublicSshKey(user);
+		if (userPublicSshKey == null || userPublicSshKey.trim().isEmpty()) {
+			return FileUtil.fileToString(getServerPublicSshKeyFilename());
+		} else {
+			return userPublicSshKey;
+		}
+	}
+
     protected String getPublicSshKey(Run run, User user) throws ValidationException, IOException {
     	if (run.getType() == RunType.Run) {
-    		return getUserPublicSshKey(user);
+			return getUserOrServerPublicSshKey(user);
     	} else {
-    		String publicSshKeyFile = getPublicSshKeyFileName(run, user);
-    		return FileUtil.fileToString(publicSshKeyFile);
+			String publicSshKeyFile = getPublicSshKeyFileName(run, user);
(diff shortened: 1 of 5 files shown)


jar-service/src/main/java/com/sixsq/slipstream/user/UserResource.java
```diff
@@ -91,16 +95,30 @@ protected boolean isMachineAllowedToAccessThisResource(){
...
+	private void replaceUserPublicSshKeyWithServerSshPublicKeyIfEmpty(User user)
 			throws ConfigurationException, ValidationException {
+
+		Cookie cookie = CookieUtils.extractAuthnCookie(getRequest());
+
+		if (CookieUtils.isMachine(cookie) == true) {
+			UserParameter userPublicSshKey = user.getParameter(ExecutionControlUserParametersFactory.CATEGORY + "."
+					+ UserParametersFactoryBase.SSHKEY_PARAMETER_NAME);
+
+			if (userPublicSshKey != null && userPublicSshKey.getValue().trim().isEmpty()) {
```
loomis: Should this be: "x == null || x.getValue()...".

jar-connector/src/main/java/com/sixsq/slipstream/connector/ConnectorBase.java
```diff
@@ -294,18 +294,31 @@ protected String getPrivateSshKeyFileName() throws ConfigurationException, Valid
...
+		return Configuration.getInstance().getProperty(ServiceConfiguration.CLOUD_CONNECTOR_ORCHESTRATOR_PUBLICSSHKEY);
+    }
+
     private String getUserPublicSshKey(User user) throws ValidationException, IOException {
 		return user.getParameter(
 				ExecutionControlUserParametersFactory.CATEGORY + "." + UserParametersFactoryBase.SSHKEY_PARAMETER_NAME)
 				.getValue();
     }
 
+    private String getUserOrServerPublicSshKey(User user) throws ValidationException, IOException {
```
loomis: Generally, is there a reason to insert the server's SSH key rather than just using a dummy value?  The problem with inserting a working key is that it looks like we're providing a backdoor into the machine.

schaubl: Yes, the reason is that some Clouds will check the format of the key and fail if the format is wrong.
2 others solutions:
- Generating a new keypair each time (need JSch as a dependency)
- Generating a keypair on my computer and then destroy the private key and hardcode the public key in SlipStream.

@loomis What do you think ?

jar-service/src/main/java/com/sixsq/slipstream/user/UserResource.java
schaubl: I don't think so because if I do that, we will enter into the `if` if `userPublicSshKey` is null and then we will get a NullPointerException because of `userPublicSshKey.setValue(serverPublicSshKey)` and `userPublicSshKey.getName()`.
@loomis Do you agree ?

jar-connector/src/main/java/com/sixsq/slipstream/connector/ConnectorBase.java
loomis: Providing something that is a valid key will have the same concerns as providing the server's public key--it looks like a backdoor has been added.  I think that we can leave it like this for now, but we should add a task to investigate how to force all of the cloud connectors to not configure SSH for the user in the case that they've not provided their own key.

schaubl: Ok, I agree.
It should be possible to not add any public key but this will require a change on all connectors.
I created an issue for that.

jar-service/src/main/java/com/sixsq/slipstream/user/UserResource.java
loomis: ok

PR #275, opened 2015-03-16
Uniqueness test made with user named 'user'

st:
So that cleanup actually deletes the row inserted by this test.
Otherwise previous execution polluted it and Continuous Integration is unstable.
@loomis @schaubl @konstan

### DIFF
--- a/jar-persistence/src/test/java/com/sixsq/slipstream/persistence/VmTest.java
+++ b/jar-persistence/src/test/java/com/sixsq/slipstream/persistence/VmTest.java
@@ -225,9 +225,9 @@ public class VmTest {
 			EntityTransaction transaction = em.getTransaction();
 			transaction.begin();
 			Query query1 = em
-					.createNativeQuery("insert into Vm values (10, 'lokal', 'instance100', null, null, 'up', 'joe')");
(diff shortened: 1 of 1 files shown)


jar-persistence/src/test/java/com/sixsq/slipstream/persistence/VmTest.java
```diff
@@ -225,9 +225,9 @@ public void cloudInstanceIdUserMustBeUnique() throws Exception {
 			EntityTransaction transaction = em.getTransaction();
 			transaction.begin();
 			Query query1 = em
-					.createNativeQuery("insert into Vm values (10, 'lokal', 'instance100', null, null, 'up', 'joe')");
+					.createNativeQuery("insert into Vm values (10, 'lokal', 'instance100', null, null, 'up', 'user')");
 			Query query2 = em
```
schaubl: @st: I think it's better to use the variable user (line 46 of this file) so if at some point the name is changed, it will not break the test.

st: Done

PR #277, opened 2015-03-17
Improves requests insertion to avoid concurrent collections

st:
A request for work is marked as busy just before it will be put in channel.
Prior this change, the busy flag occurred in the worker itself (too late). So that the online collector and offline collector could end up working on the same pair (user/connector).

@konstan I let you double check.


st:
aitäh

PR #278, opened 2015-03-17
Event resource

st:
This big pull requests brings the Clojure REST resources (based on Cal's work for CIMI) into SlipStream. 
* First use from Java world is with the Event resource, that receives messages from Run resource  (creation, state change, abort, scaling).
* This Event resource is currently internal (redirection proxy in Restlet was not included as authentication still missing).
* The persistence for Clojure resources uses the same database as Java server.

@loomis @schaubl @mebster @konstan @rbf


schaubl:
I will have a look to this pull request.

schaubl:
The Java part looks ok.
The following files has no "real" change, so they may be excluded from this pull request:
- RootApplication.java (https://github.com/slipstream/SlipStreamServer/pull/278/files#diff-be0e31281a464eedd440e785b644a372
- UserListResource.java (https://github.com/slipstream/SlipStreamServer/pull/278/files#diff-473bc3276f0b7856f58d481ac4a6df83)
Java dependencies may be moved to the SlipStream parent.

I will let somebody else do the review of the Clojure part @loomis ? @rbf ?

st:
Following @schaubl advice, files with no real changes were reverted back (https://github.com/slipstream/SlipStreamServer/commit/8ff29f2bbb399fde4e483fd7c626121772e380cd).

ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/db/impl.clj
```diff
@@ -0,0 +1,26 @@
+(ns com.sixsq.slipstream.ssclj.db.impl
+  (:require
+    [com.sixsq.slipstream.ssclj.db.binding :as p]
+    [clojure.tools.logging :as log]))
+
+(def ^:dynamic *impl*)
+
+(defn set-impl!
+  [impl]
+  (alter-var-root #'*impl* (constantly impl)))
```
rbf: @loomis 
I don't have the full overview of this code and therefore I wonder why a `binding` form or a `(def impl (delay...` form are not used instead of changing the value of `*impl*` with `alter-var-root`, which will affect the value on all threads.

loomis: The variable probably could be bound from the top-level ring application, but this is only changed once when the application starts based on values read in from the service configuration.

PR #303, opened 2015-03-30
Add extra fields to the /vms resource

### LINKED ISSUE #281: Add extra fields to the /vms resource
- [x] IP
- [ ] <del>Node instance name</del>
   - [x] Name
   - [x] Node name
   - [x] Node instance id

Fix #242.

### LINKED ISSUE #297: Bug with vmstate in Runs - vmstate may stays in Unknown even if the VM is known
This bug appear usually for orchestrators.
It probably happens when SlipStream is aware of `the link between the Run and the VM` after it sees the VM for the first time. This happens when the `cloud-run-instance` takes time to return.

### LINKED ISSUE #305: Add the possibility to filter by `Run owner` and by `User`
Filter by `User` is allowed only for privileged users.

schaubl:
Added the following extra fields to the /vms resource:
- name
- nodeName
- nodeInstanceId
- ip
- runOwner

Added the possibility to filter by `Run owner` and by `User`.
`User` is reserved for privileged users.

Fix #281 and fix #305.
Also fix #297.

Corresponding issue on the UI: slipstream/SlipStreamUI#254.


jar-persistence/src/main/java/com/sixsq/slipstream/persistence/Vm.java
```diff
@@ -87,6 +87,18 @@
...
+
+	@Attribute(required = false)
+	private String name;
+
+	@Attribute(required = false)
+	private String nodeName;
+
+	@Attribute(required = false)
+	private String nodeInstanceId;
+
```
konstan: Will DB migration be required? 
Will the attributes be rendered in the document by the serialiser for the old records that don't have values for the fields?

schaubl: > Will DB migration be required?

No


> Will the attributes be rendered in the document by the serialiser for the old records that don't have values for the fields?

No

konstan: OK.

schaubl:
@konstan : Can you have a look to my last two commits ?

PR #307, opened 2015-04-01
Add not require notes

mebster:


jar-persistence/src/main/java/com/sixsq/slipstream/persistence/Node.java
```diff
@@ -223,66 +221,57 @@ public void setName(String name) {
 	 * Assembled notes. Includes notes from inherited images.
 	 */
 	@Transient
-	@ElementArray(required = false)
+	@ElementArray(required = false, entry = "note")
 	public String[] getNotes() {
```
schaubl: @mebster 
The entry should be `notes` and not `node`.

schaubl: `entry` is for the item.
So there is no issue.

PR #311, opened 2015-04-08
Usage

st:
First version of recording of usage per user.
The only persisted metric is the number of vm.


war/pom.xml
```diff
@@ -38,13 +38,13 @@
...
 -->
 
-<!--
     <dependency>
       <groupId>com.sixsq.slipstream</groupId>
       <artifactId>SlipStreamConnector-CloudStack-jar</artifactId>
       <version>${project.version}</version>
     </dependency>
 
+<!--
```
rbf: @st 
Did you un-outcomment (?!) the CloudStack connector `<dependency>` in the `war/pom.xml` on purpose?

rbf:
This morning we reviewed the clojure parts of this PR with @st. He has shown me the main components and structures and nothing seems shocking for me. Tests coverage seems reasonable and it allowed us to refactor some code while doing the review.

@loomis 
This is quite a big PR. IMO it makes sense that somebody else has a look on the Java side as well before merging.

war/pom.xml
st: good catch.
meant for local test only.

jar-connector/src/test/java/com/sixsq/slipstream/persistence/VmTest.java
```diff
@@ -222,6 +224,7 @@ public void empty() throws ConfigurationException, ValidationException {
 	}
 
 	@Test
+	@Ignore
```
loomis: Is this test no longer relevant?  What is the reason for ignoring it now?

jar-connector/src/test/java/com/sixsq/slipstream/persistence/VmTest.java
```diff
@@ -36,8 +36,10 @@
```
loomis: The tests have moved from the `java-persistence` module to the `java-connector` module. This means that `Vm.java` and `VmTest.java` are in completely different modules, which isn't a good practice.  I suspect this has been done to avoid a circular dependency between these modules.  However, rather than splitting tests would it be a better idea to move the Collector (which depends on persistence and the connectors) to a "neutral" location?

st: I would rather split tests :
* those specific to Vm stay in VmTest (but moved to proper project : Persistence)
* those using Collector in a new CollectorTest (same place as current VmTest).

loomis: Splitting the relevant tests into the associated modules would be fine.

PR #314, opened 2015-04-11
Revert "Added the image parameter 'platform' as a RuntimeParameter. #312"

schaubl:
This reverts commit 2f8becb10bdacadec2aa6c2196caa7724af8dd56 of PR #312.

It's not needed to add the platform as a RuntimeParameter to make Windows deployment  working because the platform is taken from the image attributes.

If you think this platform is useful as a RuntimeParameter (for example to be used with `ss-get` by a deployment script), you can close this PR.
If you think it's useless, you can merge this PR.

Connected to slipstream/SlipStreamClient#89


mebster:
After discussion, having the platform in the runtime parameters can be interesting for scripts. We therefore keep it.

PR #327, opened 2015-04-28
Allow Connectors to define if a VM is usable or not 

schaubl:
This is used by the `usage` and the `metering` section of the dashboard.
It's also used by the `/usage` resource.

@st: Can you check the way the `UsageRecorder` is used in `Collector.java` ?

Connected to slipstream/slipstream#52.


st:
@schaubl UsageRecorder use looks good to me.

PR #328, opened 2015-04-28
Feature/usage resource

st:
Connected to sixsq/tasklist#226

### DIFF
--- a/ssclj/jar/pom.xml
+++ b/ssclj/jar/pom.xml
@@ -11,7 +11,7 @@
   <parent>
     <groupId>com.sixsq.slipstream</groupId>
     <artifactId>SlipStreamCljResources</artifactId>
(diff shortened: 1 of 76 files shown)


ssclj/jar/pom.xml
```diff
@@ -163,6 +163,11 @@
       <artifactId>expectations</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>honeysql</groupId>
+      <artifactId>honeysql</artifactId>
+      <version>0.5.2</version>
```
loomis: Please remove the explicit version from here.  I've added it to the main `pom.xml` file and pushed this into master.  Are `korma` and `honeysql` complementary or would it make sense to phase out `korma` use?

Issue #324, opened 2015-04-27
Add pagination to the /event resource


schaubl:
E.g: `?offset=200&limit=100`
Data returned should contain the following elements:
```cfg
offset=200 
limit=100
count=43
totalCount=243
```

- [x] Implement a CIMI compliant equivalent
- [x] Do the conversion from SS representation and CIMI equivalent 

Related to: 
- #323
- #325

loomis:
The CIMI parameters associated with paging are `$first` and `$last`.  The description can be found in Section 4.1.6.2 of the [CIMI Specification](http://dmtf.org/sites/default/files/standards/documents/DSP0263_1.1.0.pdf).  The value of the `count` attribute is defined by the standard and, I believe, corresponds to the count of items returned in the collection.  The `totalCount` could be added as a property of the collection.

I vaguely remember seeing somewhere in the specification that there can be a default value for `$last`, essentially providing a limit on the number of values returned in the collection.  However, I can't find this reference in the text at the moment.

st:
@schaubl Organise ourselves so that we pair-program while I'm in the office.

st:
- [ ] re-use code to validate offset, limit in ssclj redirector

PR #343, opened 2015-05-08
ssclj pagination

schaubl:
Connected to #324.

### DIFF
--- a/ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/db/database_binding.clj
+++ b/ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/db/database_binding.clj
@@ -1,18 +1,18 @@
 ;; database implementation of Binding protocol
 (ns com.sixsq.slipstream.ssclj.db.database-binding
   (:refer-clojure :exclude [update])
-  (:require 
-    [clojure.java.jdbc :refer :all :as jdbc]    
-    [com.sixsq.slipstream.ssclj.db.binding :refer [Binding]]
+  (:require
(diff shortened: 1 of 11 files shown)


ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/db/database_binding.clj
```diff
@@ -52,56 +52,62 @@
...
-
-(defn- delete-resource 
+  [collection-id options]
+  (let [{:keys [offset limit]} (u/offset-limit options)]
+    (if (neg? limit)
+      []
+      (->> (select resources
+                   (where {:id [like (str collection-id"%")]})
+                   (kc/limit offset)
+                   (kc/offset limit))
```
rbf: @schaubl @st 
This [two lines](https://github.com/slipstream/SlipStreamServer/blob/issue_%23324-ssclj_pagination/ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/db/database_binding.clj#L67-L68) look suspicious.
We had a look with @schaubl without understanding why it works anyway.
I won't merge it until it's fixed or understood.
Cheers.

st: Well spotted @rbf 
These 2 lines are actually wrong. The thing is that the new pagination (CIMI like with $first and $last) was well tested end-to-end for the other resource using (`usage`), but not for `event`. 
The highlighted lines concern the general implementation of `find-resources` that apply for `event` resource.

Issue #353, opened 2015-05-17
The serialization of UserParameter has changed and doesn't work with the UI


schaubl:
The server started to serialize `UserParameter` like the following:
```xml
<entry>
    <string>Exoscale.username</string>
    <userParameter category="Exoscale" description="Exoscale account username" mandatory="true" name="Exoscale.username" order="10" order_="10" readonly="false" type="RestrictedString">
        <value/>
    </userParameter>
</entry>
```

instead of: 
```xml
<entry>
    <string>Exoscale.username</string>
    <parameter category="Exoscale" description="Exoscale account username" mandatory="true" name="Exoscale.username" order="10" order_="10" readonly="false" type="RestrictedString">
        <value/>
    </parameter>
</entry>
```

`<parmeter>` is now `<UserParameter>`.
Because of that, the UI is not able to see any UserParameter.

schaubl:
I deployed a new SlipStream from `snapshot` and it had this issue. 
So this has to be fixed before the release.

schaubl:
I haven't found a reason in the code for this change.

I think the reason is the update of `simple-xml` from `2.2.4` to `2.7.1`.
See https://github.com/slipstream/SlipStream/blame/master/pom.xml#L503.

So in this case the UI will have to be updated.

schaubl:
I confirm that the cause is a side effect of the migration from `simple-xml 2.2.4` to `simple-xml 2.7.1`.

mebster:
This is easy to fix by specifying the name the element should have using @ElementList or @ElementArray attribute.

schaubl:
@mebster : It's was already fixed here: #354.
I wasn't able to fix it with a parameter of @ElementMap.

PR #354, opened 2015-05-17
The serialization of UserParameter has changed and doesn't work with the UI

schaubl:
Connected to #353


loomis:
I've reverted the version of simple-xml for now.  The version can be bumped up to v2.7.1 again, with this patch applied for the next sprint.

loomis:
This was bumped back to the latest version for the v2.9 release as the associated PR was merged.

Issue #358, opened 2015-05-18
Extend POST on RTP resource to trigger state transition if RTP is ss:state


konstan:
Extend POST on RTP resource to handle the case when the POST is done on ss:state RTP. Condition to trigger the state transition: run is mutable and state is Ready. Respond with 409 if condition is not met.

https://github.com/slipstream/SlipStreamServer/blob/master/jar-service/src/main/java/com/sixsq/slipstream/run/RuntimeParameterResource.java#L173

konstan:
Instead ended up extending PUT on RTP resource if RT is ss:state.  Semantically PUT suites better for requesting the state change.

Issue #365, opened 2015-05-21
Update configuration->target/source to 1.8 in maven-compiler-plugin


konstan:

konstan:
Extract the version as a property to the parent pom.

loomis:
Plugin configuration from the parent pom is merged into local configurations of the plugin. So actually the entire configuration should be extracted to the parent pom in the "pluginManagement" section.  The explicit setting of these values everywhere else (assuming they're not being overridden locally) can be removed.

loomis:
Also reduce the dependencies in the packages to require only the JRE (headless) and not a full JDK.  This should reduce the installation time somewhat.

schaubl:
The JRE headless will reduce a lot the size&time of the Java installation.
:+1:

PR #366, opened 2015-05-22
Issue #358 put on ss state

konstan:
Extended generic PUT on RTP with state transition from Ready to Provisioning in case the RT is ss:state.  Prerequisites: run should be `mutable`, current state should be `Ready`, requested state should be `Provisioning`.

Connected to #358


jar-service/src/main/java/com/sixsq/slipstream/run/RuntimeParameterResource.java
```diff
@@ -218,4 +262,14 @@ private void logTimeDiff(String msg, long before, long after) {
 	protected void logTimeDiff(String msg, long before) {
 		logTimeDiff(msg, before, System.currentTimeMillis());
 	}
+
+	private boolean isGlobalStateParameter() { return RuntimeParameter.GLOBAL_STATE_KEY.equals(key); }
```
schaubl: I think a new line is welcome here :)

konstan: I recall IntelliJ separated, but I recall reverting it for some reason :-/.  I'm separating onto different lines now.  Thanks.

jar-service/src/main/java/com/sixsq/slipstream/run/RuntimeParameterResource.java
```diff
@@ -145,6 +145,9 @@ public void update(Representation entity) throws ResourceException,
 				abortOrReset(value, em);
 				setValue(value);
 			}
+		} else if (isGlobalStateParameter()) {
+			States newState = attemptChangeGlobalState(value);
+			setValue(newState.toString());
```
schaubl: With this change, we cannot force the State anymore.
The ability to force the state already helped us in the past to unblock some particular Run for some customers.
So I would prefer if we can keep this ability. Maybe by adding the following code between the line 190 and 191:
```java
} else if (getUser().isSuper()) {
    newState = toNewState;
```

jar-service/pom.xml
```diff
@@ -200,8 +200,8 @@
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-compiler-plugin</artifactId>
 				<configuration>
-					<target>1.6</target>
-					<source>1.6</source>
+					<target>1.7</target>
+					<source>1.7</source>
```
schaubl: It should be updated to 1.7 everywhere or defined on the top level pom if possible.
Out of curiosity, what did you do which require 1.7 ?

jar-service/src/main/java/com/sixsq/slipstream/run/RuntimeParameterResource.java
```diff
@@ -169,23 +172,64 @@ private String extractValueFromEntity(Representation entity) {
...
+	}
+
 	@Post
-	public void completeCurrentNodeState(Representation entity) {
+	public void completeCurrentNodeStateOrChangeGlobalState(Representation entity) {
 
 		String nodeName = runtimeParameter.getNodeName();
-		States newState = null;
-
+		States newState;
```
schaubl: Does this change come from an hint of Intellij ? 
I would suggest the follwing:
```java
States newState = attemptCompleteCurrentNodeState(nodeName);
```

jar-service/pom.xml
konstan: I initially used multiple-catch statements, but then when I was testing and running the server in Eclipse (yes ... still in Eclipse :)), this part of the code was failing in the JVM, so I reverted back to separate catches.  Here is the corresponding issue to change it everywhere #365

schaubl: Ok. I did a grep into the whole code and I only found 
```xml
<source>1.6</source>
```
in `SlipStreamServer/jar-service/pom.xml` 
and in `SlipStreamServer/static-content/target/results/META-INF/maven/org.clojure/clojure/pom.xml`

konstan: #365 wasn't worked on yet.  After fixing it, the source/target versions should be defined in the parent pom as properties.

jar-service/src/main/java/com/sixsq/slipstream/run/RuntimeParameterResource.java
konstan: Agreed.  Adding.

jar-service/src/main/java/com/sixsq/slipstream/run/RuntimeParameterResource.java
konstan: Overlooked to update it after a refactor.  Updating.  Thanks.

Issue #369, opened 2015-05-22
monitoring with sentry causes server to crash


loomis:
When starting SlipStream using a configuration that logs to sentry through raven, the server crashes with the following stack trace:
```
Caused by: java.lang.NoClassDefFoundError: com/fasterxml/jackson/core/JsonFactory
        at net.kencochrane.raven.marshaller.json.JsonMarshaller.<init>(JsonMarshaller.java:94)
```
The versions of raven and dependencies have been upgraded between v2.8 and v2.9, but it is odd that this provokes a classpath issue.

loomis:
The problem can be corrected by putting raven and all of its dependencies into the `/opt/slipstream/server/lib/logging` subdirectory.  The package currently only puts raven and the `jackson-databind` dependency.  This doesn't explain why this behavior changed between v2.8 and v2.9; more investigation is needed for that. 

In the meantime, the following should be corrected:

 - [ ] Guava version should not be hardcoded in the RPM creation
 - [ ] All of the jackson dependencies should be added to `lib/logging` in the RPM.
 - [ ] These dependencies should be removed from `SlipStreamServer/jar-service` and hence the war file.

loomis:
Connected to #370.

PR #370, opened 2015-05-22
Bug/#369 sentry logging crash

loomis:
This pull request reorganizes the logging dependencies to put everything into the `lib/logging` subdirectory of Jetty.  They have also been removed from the war file to avoid conflicts between versions.  The guava dependency version has also been changed to use the version in the main `pom.xml` file everywhere.


loomis:
Connected to #369.

Issue #359, opened 2015-05-19
Add pre- and post-scale action scripts


konstan:
Required to
- prepare to the scaling action
- recover from possible reboot and/or pick up the change (e.g., attached device)

**UPDATE:** The scripts will be documented as part of slipstream/SlipStreamClient#114

Two common pre-/post- scripts with the following skeleton
```shell
case $ACTION in
  vm_remove) ;;
  vm_resize) ;;
  disk_attach) ;;
  disk_detach) ;;
esac
```
Add the same for Windows.

PR #371, opened 2015-05-26
Minor (added pre/post scale targets definitions).

konstan:
Added definitions for pre/post scale targets persistence.

The scripts themselves will be documented as part of slipstream/SlipStreamClient#114  (this follows the same pattern as was done for OnVmAdd/Remove).

Connected to #359

### DIFF
--- a/jar-service/src/main/java/com/sixsq/slipstream/module/ImageFormProcessor.java
+++ b/jar-service/src/main/java/com/sixsq/slipstream/module/ImageFormProcessor.java
@@ -180,7 +180,7 @@ public class ImageFormProcessor extends ModuleFormProcessor {
 
 		Set<Target> targets = new HashSet<Target>();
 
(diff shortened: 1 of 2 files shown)


jar-service/src/main/java/com/sixsq/slipstream/module/ImageFormProcessor.java
```diff
@@ -180,7 +180,7 @@ private void parseTargets(Form form) throws ValidationException {
 
 		Set<Target> targets = new HashSet<Target>();
 
-		for (String targetName : TARGET_SCRIPT_NAMES) {
```
schaubl: I think the definition of `TARGET_SCRIPT_NAMES` should be removed (it's at the top of this file).

schaubl: Done in 20a9b4f5686ff4eecc1d93f2781d476979a06479.

PR #374, opened 2015-06-01
Feature/resource filtering

st:
Please, at your convenience, review this pull request.
@loomis @rbf


loomis:
I did some further checks to test out your change with dropping the :Filter.  That was fine, but I found another bug in which the filter "a=1 and b=2" and combining two filters of "a=1" and "b=2" did not provide the same AST.  I've corrected that in the commit 9b27a88.  You should check that this change in the AST shape doesn't affect your stuff.  (It shouldn't but one never knows.)  Everything else looks fine.  So just merge the PR once you've looked over my changes.

PR #399, opened 2015-07-01
Feature/api naming

st:


jar-service/src/main/java/com/sixsq/slipstream/application/RootApplication.java
```diff
@@ -387,7 +387,7 @@ private void attachMetering(RootRouter router) throws ConfigurationException, Va
 
 	private void attachSSCLJ(RootRouter router) throws ValidationException {
 		for (String sscljResourceName : SSCLJRouter.SSCLJ_RESOURCE_NAMES) {			
-			guardAndAttach(router, new SSCLJRouter(getContext(), sscljResourceName), sscljResourceName);			
+			guardAndAttach(router, new SSCLJRouter(getContext(), sscljResourceName), "api/" + sscljResourceName);
```
loomis: Couple of questions on this code:
  * Why don't you just have a single router for everything below "api/"?  As it is now, we will have to modify the java side of the code every time a new resource is created and this doesn't expose the "api/" resource. 
  * It looks like the implementation still uses camel-cased URLs on the the java side.  Is this correct?

st: 1) yes, this can be improved. 
2) this is a good catch, but for the other pull-request.

PR #413, opened 2015-07-10
Feature/firewall resource

st:


ssclj/jar/src/test/clojure/com/sixsq/slipstream/ssclj/resources/lifecycle_test_utils.clj
```diff
@@ -27,9 +28,17 @@
 
 (defn is-key-value [m k v]
   (let [actual (get-in m [:response :body k])]
+    (when-not (= v actual)
+      (println "Expecting " v " got " actual))
```
rbf: @st Is that not already done by the `is` itself?

ssclj/jar/src/test/clojure/com/sixsq/slipstream/ssclj/resources/network_service_test.clj
```diff
@@ -0,0 +1,159 @@
+(ns com.sixsq.slipstream.ssclj.resources.network-service-test
+  (:require
+    [clojure.test :refer :all]
+    [peridot.core :refer :all]
+
+    [korma.core :as kc]
+
+    [com.sixsq.slipstream.ssclj.api.acl :as acl]
```
rbf: @st [minor] In some files above `require`s are vertically aligned, and here they aren't. Personally, I like it vertically aligned, but anyway consistency should be present.

rbf:
Looks ok for me.
Merging.

PR #418, opened 2015-07-13
Fix #405

### LINKED ISSUE #405: Events JSON is not consistently sorted by 'timestamp'.
It's expected to be sorted from newest to oldest by the `timestamp` value. See https://nuv.la/event (not as `super`!)

st:
The collection (result of query) is ordered specifically by each
resource, just before pagination.
(default ordering is to simply return the collection).
Event collection sorts the collection by :timestamp desc.

### DIFF
--- a/ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/resources/event.clj
+++ b/ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/resources/event.clj
@@ -1,14 +1,14 @@
 (ns 
   com.sixsq.slipstream.ssclj.resources.event
   (:require
-    [clojure.tools.logging :as log]
-    [schema.core :as s]
-    [com.sixsq.slipstream.ssclj.resources.common.authz :as a]
-    [com.sixsq.slipstream.ssclj.resources.common.crud :as crud]
-    [com.sixsq.slipstream.ssclj.resources.common.std-crud :as std-crud]
-    [com.sixsq.slipstream.ssclj.resources.common.utils :as u]
-    [com.sixsq.slipstream.ssclj.resources.common.schema :as c]
-    [com.sixsq.slipstream.ssclj.resources.common.debug-utils :as du]))
+    [clojure.tools.logging                                    :as log]
+    [schema.core                                              :as s]
+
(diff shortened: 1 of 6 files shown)


ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/resources/event.clj
```diff
@@ -1,13 +1,14 @@
...
-    [com.sixsq.slipstream.ssclj.resources.common.utils      :as u]
-    [com.sixsq.slipstream.ssclj.resources.common.schema     :as c]))
+    [clojure.tools.logging :as log]
+    [schema.core :as s]
+    [com.sixsq.slipstream.ssclj.resources.common.authz :as a]
+    [com.sixsq.slipstream.ssclj.resources.common.crud :as crud]
+    [com.sixsq.slipstream.ssclj.resources.common.std-crud :as std-crud]
+    [com.sixsq.slipstream.ssclj.resources.common.utils :as u]
+    [com.sixsq.slipstream.ssclj.resources.common.schema :as c]
+    [com.sixsq.slipstream.ssclj.resources.common.debug-utils :as du]))
```
rbf: `require`s got unaligned again (by auto-request feature in IntelliJ, I guess)

ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/resources/event.clj
```diff
@@ -99,7 +100,21 @@
...
+  [e1 e2]
+  (compare (:timestamp e2) (:timestamp e1)))
+
+(defn sort-by-timestamps
+  [events]
+  (sort compare-timestamps-desc events))
+
+(defmethod crud/sort-collection resource-name
+  [request collection]
+  (sort-by-timestamps collection))
```
rbf: Couldn't this line be written as:

``` clojure
(sort-by :timestamp coll)
```

so that both helper functions `compare-timestamps-desc` and `sort-by-timestamps` aren't needed?

rbf:
@st 
I found nothing blocking.
Do you want to have a look at my comment above before I merge?

PR #425, opened 2015-07-17
Minor: improve logs.

st:


ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/resources/common/std_crud.clj
```diff
@@ -3,18 +3,27 @@
...
 
     [com.sixsq.slipstream.ssclj.resources.common.cimi-filter  :as cf]
     [com.sixsq.slipstream.ssclj.resources.common.pagination   :as pg]))
 
+(defn log-request
+  [request]
+  (log/info (:request-method request)
+            (:uri request)
+            (get request :query-string "")
+            (get request :body "")))
```
rbf: @st Wondering if it's more useful to explicitly log when something was not found, instead of a blank string:

``` clojure
(get request :query-string "<no :query-string>")
(get request :body         "<no :body>")
```

st: good catch.
it's even worse than that, as the keys are present, but values are nil.
so default value is useless. will remove it.

rbf: Ok. Just for reference, in such cases where `nil` would be returned since the key is present, I tend to do something like the following:

``` clojure
(-> request :query-string (or "<no :query-string>"))
```

st: Good to know.
thanks.

Issue #454, opened 2015-08-18
changes for cloud entry point


loomis:
The cloud entry point is appearing on the SlipStream server at the `api/cloud-entry-point` relative URL.  This is correct, but there are several points in the resulting resource that are not correct and need to be fixed.

 - [x] The resourceURI must be the one defined by the CIMI specification (in the DMTF namespace rather than SixSq).
 - [x] The baseURI value must be the one for the public server rather than the ssclj server on localhost, i.e. `https://nuv.la/api/` for the SaaS.
 - [x] The "id" must match the relative URL for the resource, so it should be "cloud-entry-point" rather than "CloudEntryPoint".
 - [x] All of the hrefs for the provided resources must be lisp-case (or kebab-case) rather than CamelCase.
 - [x] The relative URL for the operations on the cloud entry point is wrong.  It is "#" and should instead be "cloud-entry-point".

loomis:
All of the points except the correct baseURI have been confirmed by running a test instance.  The baseURI is still pointing to the local instance.  I'll have to verify if the headers are being properly transmitted and interpreted by the ssclj server.

loomis:
The issue with getting the forwarded host information through to the clojure server has been resolved.  The solution was to reuse the existing utilities for finding the base URL for redirection.  All of the points in the issue description have been addressed.

PR #456, opened 2015-08-19
correct issues with cloud entry point

loomis:
Connected to #454


konstan:
IMHO, looks OK.  Merging.

Issue #461, opened 2015-08-30
permit 'plugin' extensions to clojure resource server


loomis:
Modify the deployment of the clojure resource server to allow resources and jar files with extensions to be dropped into place.  This requires the following changes:
 - [x] change the current `etc` location to `resources` to make it clear that these are java resource files looked up on the classpath
 - [x] add a `lib/ext` directory in which additional jar files can be placed

It would also be useful to:
 - [x] add systemd startup configuration to allow native startup (rather than via `init.d` script) on CentOS 7

loomis:
The pull request should address all of the points raised in the issue.  In addition, information about the java vendor, version, and classpath are logged when the server starts to help with debugging of issues.

PR #462, opened 2015-08-30
allow 'plugin' extensions to clojure resource server 

loomis:
Connected to #461.

### DIFF
--- a/ssclj/rpm/src/main/systemd/ssclj.service
+++ b/ssclj/rpm/src/main/systemd/ssclj.service
@@ -1,13 +1,25 @@
 [Unit]
 Description=SlipStream API
-After=syslog.target network.target
+After=syslog.target
+After=network.target
 
 [Service]
-RemainAfterExit=yes
-Type=oneshot
-ExecStart=/etc/init.d/ssclj start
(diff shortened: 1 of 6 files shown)


ssclj/rpm/src/main/systemd/ssclj.service
```diff
@@ -1,13 +1,22 @@
 [Unit]
 Description=SlipStream API
-After=syslog.target network.target
+After=syslog.target
+After=network.target
```
st: Is it equivalent to previous version?

st: I have another question regarding `create-daily-usage.sh`
should `-cp /opt/slipstream/ssclj/etc` be updated?

loomis: Yes, the "After" configurations are equivalent and I think that the one with separated values is clearer.

loomis: I'll update the classpath for the script.

loomis:
The classpath for the script was updated to match the new layout of the server.  Please check that it is correct.

Issue #416, opened 2015-07-13
Server upgrade requiring new version of java should either remove old version or update java definition in alternatives


konstan:
Otherwise, server built against newer java fails to start with an older one.  Options:
- remove old java (as we assume we own the box)
- update `alternatives --set java /usr/lib/jvm/java-1.8.0-openjdk-1.8.0.45-28.b13.el6_6.x86_64/jre/bin/java`
- discover new java path and hardcode this in jetty's conf in RPM post-install scriplet.

konstan:
Make server RPM conflict with `java-1.7.0-openjdk`  and let sysadmin to resolve the issue.

schaubl:
With @0xbase12 we discovered that the option which was chosen conflict with the installation of the NuvlaBox.
Currently we cannot install the last SlipStream release on the NuvlaBox because StratusLab require java-1.7.0-openjdk and slipstream-server explicitly refuse to install (see 9ddb3dc0e54c506ca20dce4b45ebd83ec1f6c34a) on a system with java-1.7.0-openjdk.

So I'm suggesting to switch to the third solution which consist of providing the full path of the java-1.8 binary to Jetty instead of `/usr/bin/java`.

konstan:
Then the solution can be as follows:

1. remove ` <conflict>java-1.7.0-openjdk</conflict>` from server RPM.
2. add `JAVA` export with the required version number of `jre`
```
[root@109-231-126-66 etc]# grep JAVA /etc/default/slipstream
export JAVA=/etc/alternatives/jre_1.8.0/bin/java
[root@109-231-126-66 etc]# 
```

Changed milestone to 2.16.

konstan:
We need to update `JAVA` in `/etc/defaults/` for `ssclj` service as well.

schaubl:
:+1:

PR #469, opened 2015-09-14
Point to java from JRE version which gets defined at the package build time

konstan:
Connected to #416

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -58,6 +58,7 @@
     <logger.level>INFO</logger.level>
     <logger.filepattern>${project.build.directory}/instance/SlipStreamServer/var/log/slipstream.%g
     </logger.filepattern>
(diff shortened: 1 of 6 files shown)


loomis:
The copy and replacement of a variable in a file can be done much more efficiently, portably, and succinctly by using the resources plugin rather than `cp`/`sed` through the exec plugin.  You should only need to move the file to the `resources` area, add the file to the list of filtered files, and change the value to replace with `${your variable}` in the file.

konstan:
Thanks!  Never used it :-/  Trying now.

konstan:
@loomis Just pushed the changes with using `resources` plugin (very convenient! :-)).  Please check.

loomis:
Looks OK to me.  Files are properly changed and included in the packages.  Merging...

Issue #476, opened 2015-09-18
clean up the main and server classes for ssclj server


loomis:
Clean up and reorganize the main and server classes for the ssclj server to put methods in correct namespaces and to make the code clearer.

PR #477, opened 2015-09-18
clean up main and server namespaces for ssclj server 

loomis:
Connected to #476.


ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/app/main.clj
```diff
@@ -1,15 +1,27 @@
...
+(def ^:const default-port 8200)
 
 (defn valid-port?
   "If the port number is valid, then returns the port itself;
-   otherwise returns nil."
+   otherwise returns false."
   [port]
-  (if (< 0 port 65536)
-    port))
+  (and (< 0 port 65536) port))
```
rbf: This will return a _cryptic_ `NPE` or `ClassCastException` if the port is `nil` or a `String`. The issue is that the exception will not be directly related with that function but deeper in the `Numbers` namespace, if we can silently return `false`.

I'd suggest to add a check as a `pre`-condition if we want to be verbose `{:pre [(integer? port)]}` or as the first condition into the existent `and` sform.

loomis: Good point.  I've just integrated this into the `parse-port` function and added tests to make sure it has reasonable behavior for all inputs.  This change was pushed directly to master (859d3a5dddc1751f0966134be753d3242a60573c).

rbf: 👍

Issue #474, opened 2015-09-17
allow initialization method for resources


loomis:
Some clojure server resources may need initialization.  An existing example is the `root` resource.

Generalize this so that the `initialize` method on a loaded resources will be called after database initialization but before the service is started.  Move the `root` initialization into the `root` resource in the refactoring.

loomis:
The initialization of the resources on server startup can be found in [`server.clj`](https://github.com/slipstream/SlipStreamServer/blob/master/ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/app/server.clj#L66).

An example for the CloudEntryPoint resource can be seen in [`root.clj`](https://github.com/slipstream/SlipStreamServer/blob/master/ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/resources/root.clj#L118).

PR #481, opened 2015-09-18
allow initialization of resources before starting server

loomis:
Connected to #474.


loomis:
@0xbase12: Please check that this implementation of the resource initialization will work for you.  In particular, the initialization will take place _before_ the server is started; this presumes that in the initialization, no calls to the server itself are done. There's also quite a bit of refactoring in anticipation of the namespace loading being used for other purposes.

Issue #472, opened 2015-09-17
allow alternate web application container


loomis:
Refactor the server namespace to allow alternate web application containers to be used.  The current http-kit should be one possibility.  The other web application to support should be Aleph.  

By supporting Aleph it may make the possible transition to yada easier.  Aleph is also entirely asynchronous, so proxing the java server from ssclj may be more efficient with Aleph compared to http-kit (to be seen!).

loomis:
Two application containers are now supported: [http-kit](http://www.http-kit.org) and [aleph](https://github.com/ztellman/aleph).  This can be seen in [`server.clj`](https://github.com/slipstream/SlipStreamServer/blob/master/ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/app/server.clj#L66) and in the files [`httpkit_container.clj`](https://github.com/slipstream/SlipStreamServer/blob/master/ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/app/httpkit_container.clj) and [`aleph_container.clj`](https://github.com/slipstream/SlipStreamServer/blob/master/ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/app/aleph_container.clj). 

If we eventually move to [yada](https://github.com/juxt/yada) as the web application framework, this will be helpful because aleph is currently the only container supported by yada.  Yada has the potential benefits of routing defined via data structures, cleaner resource interface/protocol, use of Prismatic Schema (which we already use) for resource validation, and automatic generation of Swagger compatible resources (for 'painless' API documentation). The primary downside at the moment is poor and/or incorrect documentation probably due mostly to the rapid development around the 1.0.0 beta.

PR #482, opened 2015-09-20
allow alternate application containers

loomis:
Allow use of the http-kit or aleph web application containers.

Connected to #472.


loomis:
Should be merged after #481.

PR #527, opened 2015-11-07
allow metrics to be pushed to graphite

loomis:
This PR allows a graphite server to be specified on startup.  If specified, then the defined metrics (JVM, Ring, etc.) will be pushed to that server.  If not specified, then the thread to push the metrics is not started. In both cases, the metrics will continue to be available from `/api/metrics` in JSON format.

### DIFF
--- a/ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/app/server.clj
+++ b/ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/app/server.clj
@@ -76,12 +76,11 @@
    (log/info "java classpath: " (System/getProperty "java.class.path"))
    (set-db-impl)
    (resources/initialize)
-   (if (= impl "httpkit")
-     (-> (create-ring-handler)
-         (httpkit/start-container port))
-     (-> (create-ring-handler)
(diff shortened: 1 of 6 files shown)


ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/app/server.clj
```diff
@@ -79,7 +80,8 @@
      (-> (create-ring-handler)
          (httpkit/start-container port))
      (-> (create-ring-handler)
-         (aleph/start-container port)))))
+         (aleph/start-container port)))
+   (graphite/start-graphite-reporter)))
```
st: The start of graphite reporter should probably be done before the start of the web container.
This modifies the return value of this function (no longer the expected and documented stop function).

loomis: Good catch.  I've moved the metrics process before the starting of the containers but after the definition of the ring handler, where the metrics are defined.  The start/stop should work correctly from the REPL now.  The previous implementation was working from the `init.d` script, but only because the stop function isn't used there.

PR #531, opened 2015-11-11
issue uploading reports to server

st:


ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/middleware/proxy_redirect.clj
```diff
@@ -147,6 +156,13 @@ re-cookie
   (log/debug "rewrite result" result)
   result))
 
+(defn- extract-body
+  [request]
+  (let [accept (get-in request [:headers "accept-encoding"])]
+    (if (and accept (.startsWith accept "gzip"))
+      (slurp-body-binary request)
+      (slurp-body request))))
+
```
loomis: Perhaps I've misunderstood something, but this section looks strange to me.

If this is the original request, then you should be looking at the "content-type" header to determine if the request's body is a `gzip` file.  The "accept" and "accept-encoding" headers would only be useful for the response.

In any case, I think that the whole test is unnecessary.  We'll only ever be uploading gzipped tarballs, so the entire function could just be replaced with a call to `(slurp-body-binary request)`.

PR #534, opened 2015-11-18
Adapt response status to resource exception

st:
To be merged with corresponding PR on UI project.


rbf:
For reference, the corresponding PR on UI project mentioned above is https://github.com/slipstream/SlipStreamUI/pull/484

Issue #554, opened 2015-11-27
Usage records kept open when SlipStream restarts


st:
When SlipStream has been restarted, and a Usage Record must be closed (on VM termination), a check is made on a static variable that prevents the Usage Record to be actually closed.

st:
2 examples found in BB were manually fixed. (insert 'end_timestamp' at the date BB was restarted).

PR #555, opened 2015-11-27
Fix: no longer rely on static state to avoid closing record

st:
Log levels in case of inconsistent transitions lowered down.
Connected to #554

### DIFF
--- a/jar-connector/src/main/java/com/sixsq/slipstream/connector/UsageRecorder.java
+++ b/jar-connector/src/main/java/com/sixsq/slipstream/connector/UsageRecorder.java
@@ -75,9 +75,8 @@ public class UsageRecorder {
 			}
 
 			if(!hasRecorded(cloud, instanceId)){
-				logger.fine("Not recorded => avoiding inserting usage record END for "
+				logger.fine("Not recorded but still inserting usage record END for "
 						+ metrics + ", " + describe(instanceId, user, cloud));
-				return;
 			}
 
 			logger.info("Inserting usage record END, metrics" + metrics + ", for " + describe(instanceId, user, cloud));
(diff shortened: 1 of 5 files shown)


jar-connector/src/main/java/com/sixsq/slipstream/connector/UsageRecorder.java
```diff
@@ -75,9 +75,8 @@ public static void insertEnd(String instanceId, String user, String cloud, List<
 			}
 
 			if(!hasRecorded(cloud, instanceId)){
-				logger.fine("Not recorded => avoiding inserting usage record END for "
+				logger.fine("Not recorded but still inserting usage record END for "
 						+ metrics + ", " + describe(instanceId, user, cloud));
-				return;
```
konstan: What happens if we really hasn't opened a recorded for `cloud, instanceId` because we have never seen this before?  I.e. the provided pair `cloud, instanceId` for inserting the end is an artefact.  So, we have an `end` but no `start`.

st: An inconsistent transition will be detected without consequences. (only a INFO log).
See `initial` -> `stop` in ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/usage/state_machine.clj

konstan: What I meant was more related to the metrics processor when it sees `end` w/o `start` (in DB).  Which can be a result of `still inserting end` even if we there is no `cloud, instanceId` in the `recordedVmInstanceIds` (`hasRecoreded`), but not due to the service restart and loss of the records in `recordedVmInstanceIds` map, but rather real absence of the `start`.

Issue #557, opened 2015-11-27
Super does not see events for all runs


st:
Super can see all runs, but the events of a specific run are always empty.

PR #558, opened 2015-11-27
Fix: pass ADMIN role to API when user is super

st:
connected to #557

### DIFF
--- a/jar-service/src/main/java/com/sixsq/slipstream/api/APIBaseResource.java
+++ b/jar-service/src/main/java/com/sixsq/slipstream/api/APIBaseResource.java
@@ -67,9 +67,16 @@ public abstract class APIBaseResource extends BaseResource {
             headers = new Series<Header>(Header.class);
             resource.getRequestAttributes().put("org.restlet.http.headers", headers);
         }
-        headers.add("slipstream-authn-info", getUser().getName());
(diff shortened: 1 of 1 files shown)


jar-service/src/main/java/com/sixsq/slipstream/api/APIBaseResource.java
```diff
@@ -67,11 +67,18 @@ private String getAPIResourceAsJSON() throws IOException {
...
+        headers.add("slipstream-authn-info", getNameRoles());
 
         return resource.get().getText();
     }
 
+    private String getNameRoles() {
+        String userName = getUser().getName();
+        String roleAdminIfSuper = getUser().isSuper() ? " ADMIN" : "";
+
+        return userName + roleAdminIfSuper;
```
schaubl: This is dangerous because we rely on a variable which contains a user input (the username) to check if it is admin.
So if a user create an account with the name `hacker ADMIN` he will get admin privileges for SSCLJ resources.
I tested this behavior and I didn't found a way to exploit this because the server refuse spaces in the username.
But this behavior may change in the future (during a refactor, code review, code change, ...)
So I suggest to pass the information about privileges in another header.

schaubl: The following issue was created #564.

Issue #580, opened 2015-12-07
After service startup the very first submission of a deployment fails with abort msg "Unauthorized"


konstan:
POST+redirect+GET succeeds and user gets the URL of the run.  But then, run fails with "Unauthorized" abort message.  This is confusing as may lead to thinking that this is authz error on IaaS. 
This is highly likely related to generation of the machine cookie.  See logs below.

NB! The second submission of the deployment succeeds.

SSCLJ logs

```
2015-12-07 10:27:00,982 INFO  - :post /auth/token [no-authn-info] ?claims=%7B%22com.sixsq.expirydate%22%3A%220%22%2C%22com.sixsq.isMachine%22%3A%22true%22%2C%22com.sixsq.runId%22%3A%22fb72a168-1efa-4e17-99cd-608652bb51f3%22%2C%22cloudservice%22%3A%22atos-es1%22%2C%22com.sixsq.identifier%22%3A%22test%22%7D&token no-body
2015-12-07 10:27:00,984 INFO  - Will build-token
2015-12-07 10:27:00,987 INFO  - Will create token for claims= {:com.sixsq.expirydate 0, :com.sixsq.isMachine true, :com.sixsq.runId fb72a168-1efa-4e17-99cd-608652bb51f3, :cloudservice atos-es1, :com.sixsq.identifier test}
2015-12-07 10:27:01,004 ERROR - exception in token creation  #error {
 :cause nil
 :via
 [{:type java.lang.NullPointerException
   :message nil
   :at [java.util.regex.Matcher getTextLength Matcher.java 1283]}]
 :trace
[[java.util.regex.Matcher getTextLength Matcher.java 1283]
  [java.util.regex.Matcher reset Matcher.java 309]
  [java.util.regex.Matcher <init> Matcher.java 229]
  [java.util.regex.Pattern matcher Pattern.java 1093]
  [java.util.regex.Pattern split Pattern.java 1206]
  [clojure.string$split invoke string.clj 224]
  [buddy.sign.jws$unsign invoke jws.clj 176]
  [com.sixsq.slipstream.auth.simple_authentication$check_token_impl invoke simple_authentication.clj 111]
  [com.sixsq.slipstream.auth.simple_authentication$create_token invoke simple_authentication.clj 125]
  [com.sixsq.slipstream.auth.simple_authentication.SimpleAuthentication token simple_authentication.clj 151]
[...]

PR #581, opened 2015-12-07
Issue bug basic auth

st:
Connected to #580

### DIFF
--- a/jar-service/src/main/java/com/sixsq/slipstream/authn/CookieAuthenticator.java
+++ b/jar-service/src/main/java/com/sixsq/slipstream/authn/CookieAuthenticator.java
@@ -98,10 +98,9 @@ public class CookieAuthenticator extends AuthenticatorBase {
 
 		logger.fine("handle valid, cookie = " + cookie);
 		String tokenInCookie = CookieUtils.tokenInCookie(cookie);
-		if(tokenInCookie!=null) {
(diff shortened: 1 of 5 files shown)


jar-service/src/main/java/com/sixsq/slipstream/authn/CookieAuthenticator.java
```diff
@@ -98,10 +98,9 @@ private boolean handleValid(Request request, Cookie cookie) {
 
 		logger.fine("handle valid, cookie = " + cookie);
 		String tokenInCookie = CookieUtils.tokenInCookie(cookie);
-		if(tokenInCookie!=null) {
+		if(tokenInCookie != null) {
 			logger.fine("handle valid, tokenInCookie = " + tokenInCookie);
-			user.setAuthnToken(tokenInCookie);
-			user = user.store();
+			user.storeAuthnToken(tokenInCookie);
```
konstan: Find a way to conditionally store the authn token during the Cookie based authentication. https://github.com/slipstream/SlipStreamServer/issues/582

konstan:
Tested on a test instance.  Works.  Merging.

loomis:
If the token is stored in the user object, can the user be logged into SlipStream on more than one machine (different IP addresses)?  What happens if the IP address changes for a user, will he be implicitly logged out and need to log in again?

Issue #593, opened 2016-01-13
Add roles to authn tokens


st:

st:
![roles-token 001](https://cloud.githubusercontent.com/assets/16004/12484683/b2999138-c05b-11e5-93e6-96caa0e94089.png)

st:
- [ ] special roles to be taken care of.

PR #608, opened 2016-01-21
Issue add roles in token

st:
Connected to #593


jar-service/src/main/java/com/sixsq/slipstream/api/APIBaseResource.java
```diff
@@ -73,10 +77,18 @@ private String getAPIResourceAsJSON() throws IOException {
...
 
-        return userName + roleAdminIfSuper;
+        List<String> nameRoles = new ArrayList<>();
+
+        nameRoles.add(getUser().getName());
+        if (getUser().isSuper()) {
+            nameRoles.add("ADMIN");
+        }
+        if (getUser().getRoles() != null) {
+            nameRoles.addAll(Arrays.asList(StringUtils.split(getUser().getRoles(), ",")));
```
loomis: This should strip out any special roles that we have.  From memory these are "ADMIN", "USER", and "ANON".

Issue #567, opened 2015-12-02
Horizontal scaling down fails on authn service side


konstan:
```
2015-12-02 14:17:49,564 INFO  - :delete /run/d775a1bb-0e2c-4a2a-af11-9ea78cfed3e3/testclient [no-authn-info] ?no-query-string java.io.ByteArrayInputStream@316a4a72
2015-12-02 14:17:49,566 ERROR - error in HTTP handler
java.lang.IllegalArgumentException: No matching method found: setEntity for class org.apache.http.client.methods.HttpDelete
        at clojure.lang.Reflector.invokeMatchingMethod(Reflector.java:53)
        at clojure.lang.Reflector.invokeInstanceMethod(Reflector.java:28)
        at puppetlabs.http.client.async$fn__9585$request_with_client__9586$fn__9587.invoke(async.clj:398)
        at puppetlabs.http.client.async$fn__9585$request_with_client__9586.invoke(async.clj:362)
        at puppetlabs.http.client.sync$request_with_client.invoke(sync.clj:29)
        at puppetlabs.http.client.sync$request.invoke(sync.clj:40)
        at puppetlabs.http.client.sync$delete.invoke(sync.clj:92)
        at com.sixsq.slipstream.ssclj.middleware.proxy_redirect$redirect.invoke(proxy_redirect.clj:108)
        at com.sixsq.slipstream.ssclj.middleware.proxy_redirect$wrap_proxy_redirect$fn__302.invoke(proxy_redirect.clj:128)
        at ring.middleware.cookies$wrap_cookies$fn__2282.invoke(cookies.clj:161)
        at com.sixsq.slipstream.ssclj.middleware.logger$wrap_logger$fn__107.invoke(logger.clj:41)
        at aleph.http.server$wrap_stream__GT_input_stream$fn__6577.invoke(server.clj:403)
        at aleph.http.server$handle_request$fn__6495$f__2914__auto____6496.invoke(server.clj:153)
        at clojure.lang.AFn.run(AFn.java:22)
        at io.aleph.dirigiste.Executor$Worker$1.run(Executor.java:49)
        at java.lang.Thread.run(Thread.java:745)
```

konstan:
Client-server exchange.

```
$ ss-node-remove -vvv -u test -p 85e56427ff41 --endpoint https://185.19.30.134 d775a1bb-0e2c-4a2a-af11-9ea78cfed3e3 testclient 1

: 2015-12-02T14:24:51Z :
[...]

PR #610, opened 2016-01-22
Do not send body when HTTP verb is DELETE

st:
See https://tickets.puppetlabs.com/browse/TK-66
Connected to #567

### DIFF
--- a/ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/middleware/proxy_redirect.clj
+++ b/ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/middleware/proxy_redirect.clj
@@ -1,6 +1,7 @@
 (ns com.sixsq.slipstream.ssclj.middleware.proxy-redirect
   (:require
     [clojure.tools.logging :as log]
+    [com.sixsq.slipstream.ssclj.middleware.logger :as ml]
     [superstring.core :as str]
 
     [puppetlabs.http.client.sync :as http]
@@ -43,7 +44,7 @@
 
 (defn- slurp-body-binary
   [request]
-  (if-let [len (when (:body request) (get-in request [:headers "content-length"]))]
+    (if-let [len (when (and (not= :delete (:request-method request)) (:body request)) (get-in request [:headers "content-length"]))]
     (-> request
         :body
         (slurp-binary (Integer/parseInt len))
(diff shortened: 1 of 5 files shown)


ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/middleware/proxy_redirect.clj
```diff
@@ -43,7 +44,7 @@
 
 (defn- slurp-body-binary
   [request]
-  (if-let [len (when (:body request) (get-in request [:headers "content-length"]))]
+    (if-let [len (when (and (not= :delete (:request-method request)) (:body request)) (get-in request [:headers "content-length"]))]
```
konstan: Maybe like this?
```
(defn delete-req?
  [request]
  (= :delete (:request-method request)))

 (defn- slurp-body-binary
    [request]
    (if-not (delete-req? request)
      (if-let [len (when (:body request) (get-in request [:headers "content-length"]))]
        ...

```

st: See refactor.

ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/middleware/proxy_redirect.clj
```diff
@@ -41,9 +42,23 @@
...
 
-(defn- slurp-body-binary
+(defn- not-delete?
   [request]
-  (if-let [len (when (:body request) (get-in request [:headers "content-length"]))]
+  (not= :delete (:request-method request)))
+
+(defn- body?
+  [request]
+  (:body request))
```
konstan: This `body?` is not a predicate in the canonical sense.  Maybe better make it to return boolean.  Something like
```
(not (nil? (:body request)))
```

st: You're right.
Can you do the change and merge?

loomis: My 2 cents.  The two expressions `(:body request)` and `(body? request)` seem equally clear and for all intents and purposes have the same effect.  Is `body?` needed at all?

st: I added this function to be symmetric with first check (not-delete?).
It seemed clearer to me.

Issue #624, opened 2016-02-15
cleanup kibit and expectation warnings


loomis:
Clean up the kibit and expectation warnings in the auth and ssclj modules.  These are largely related to overuse of the threading operators and use of deprecated functions from prismatic schema.

PR #625, opened 2016-02-15
Feature/static analysis fixes

loomis:
Connected to #624.


loomis:
As part of the PR I've reformatted all of the clojure source files with IntelliJ.  I've used the default parameters plus the option to format let bindings.  The let binding formatting aligns values, unless destructuring is used; don't know why this isn't handled by the formatter.

In any case, we can discuss the formatting rules we want to follow.  My only unbendable requirement is that the set of rules must allow for complete, correct auto-formatting of the source files without any need for manual changes.

Issue #643, opened 2016-02-24
Changes done outside of SlipStream prevent close of usage records


st:
If the instance type of a VM is changed *outside* of SlipStream, we are currently unable to close the previous value.
For example, the VM for that run (https://nuv.la/run/cd8f5a49-820c-4f79-9669-0a3b2ac8cc79#slipstream+slipstream-1) was changed (instance type) in exoscale dashboard.
The corresponding usage record was left open, even after the VM was terminated.

konstan:
What about cpu/ram(/disk size?) metrics that can change during the lifetime of VM same way as instance-type, i.e. "outside" of SlipStream?  I'm not sure this is covered.

PR #662, opened 2016-03-15
Issue #643 fix close usage records

st:
Connected to #643


konstan:
What about cpu/ram(/disk size?) metrics that can change during the lifetime of VM same way as instance-type, i.e. "outside" of SlipStream? I'm not sure this is covered.

Issue #644, opened 2016-02-24
Avoid noise in usage records when manually changing an instance id


st:
The current implementation tries to close usage records already closed, resulting in:
* unnecessary HTTP calls from Web Server to API server
* noise in API server (lots of `Action :stop  is not allowed for state  :stopped` messages)

PR #663, opened 2016-03-16
Prevent useless re-closing of usage records

st:
Also removes useless logging
Connected to #644


jar-connector/src/main/java/com/sixsq/slipstream/connector/UsageRecorder.java
```diff
@@ -1,9 +1,18 @@
 package com.sixsq.slipstream.connector;
 
 import com.sixsq.slipstream.persistence.Vm;
-
-import java.util.*;
+import org.bouncycastle.math.ec.ScaleYPointMap;
```
konstan: Unused import.

Issue #565, opened 2015-12-02
Horizontal scale up fails on system-testing deployment for testclient with 0 initial multiplicity


konstan:
Scaling up `apache` component from 0 multiplicity to 1 works, but fails for the `testclient` component.

```
...
Caused by: java.lang.NullPointerException
        at com.sixsq.slipstream.factory.DeploymentFactory.addParameterMapping(DeploymentFactory.java:273)
        at com.sixsq.slipstream.run.RunNodeResource.createNodeInstanceRuntimeParameters(RunNodeResource.java:292)
        at com.sixsq.slipstream.run.RunNodeResource.createNodeInstanceOnRun(RunNodeResource.java:276)
        at com.sixsq.slipstream.run.RunNodeResource.addNodeInstancesInTransaction(RunNodeResource.java:142)
        at com.sixsq.slipstream.run.RunNodeResource.addNodeInstances(RunNodeResource.java:113)
        ... 79 more
2015-12-02T10:57:21.328+0000 INFO org.restlet.engine.log.LogFilter afterHandle 2015-12-02       10:57:21        127.0.0.1       test    127.0.0.1       8182    POST    /run/d775a1bb-0e2c-4a2a-af11-9ea78cfed3e3/testclient    -       500     339     -       80      https://localhost:8182  Python-httplib2/0.9.2 (gzip)    -
...
 2015-12-02T10:57:29.889+0000 INFO org.restlet.engine.log.LogFilter afterHandle 2015-12-02      10:57:29        127.0.0.1       test    127.0.0.1       8182    POST    /run/d775a1bb-0e2c-4a2a-af11-9ea78cfed3e3/apache        -       201     8       -       215     http://localhost:8182   Python-httplib2/0.9.2 (gzip)    -
```

konstan:
The scaling up (0 -> 1) of `testclient` succeeds after scaling up of `apache`.

konstan:
The issue is not a regression in v2.19.2-SNAPSHOT.  It's in v2.19.1 as well.  Tested on https://bb1.hnx.helix-nebula.eu/run/95815b10-c5e0-498b-8ade-1e424b7f2c67

PR #682, opened 2016-03-30
avoid NPE if the multiplicity of the output parameter provider is zero i.e. mapping is not present

konstan:
Connected to #565


jar-connector/src/main/java/com/sixsq/slipstream/factory/DeploymentFactory.java
```diff
@@ -20,32 +20,20 @@
...
-import com.sixsq.slipstream.persistence.Node;
-import com.sixsq.slipstream.persistence.NodeParameter;
-import com.sixsq.slipstream.persistence.Parameter;
-import com.sixsq.slipstream.persistence.ParameterCategory;
-import com.sixsq.slipstream.persistence.Run;
-import com.sixsq.slipstream.persistence.RunParameter;
-import com.sixsq.slipstream.persistence.RunType;
-import com.sixsq.slipstream.persistence.RuntimeParameter;
-import com.sixsq.slipstream.persistence.User;
+import com.sixsq.slipstream.persistence.*;
```
loomis: It would be better to avoid wildcard imports.

konstan:
@loomis please check.  Removed wildcard imports.  
(Updated configuration of my IDE.)

PR #717, opened 2016-05-16
change delete response from 204 to 200

loomis:


loomis:
Pull request SixSq/SlipStreamServer#12 must be merged at the same time.

Issue #694, opened 2016-04-20
Migrate resources data from base to elastic search


st:
A script should transfer content of database to elastic search store for all resources:
- [x] bulk insert
- [x] modify each ID
- [x] event
- [x] usage-record
- [x] usage-summary
- [ ] work in batch for insert (count, partition, and use limit/offset) 
- [ ] service-offer
- [ ] service-attribute
- [ ] deploy SS, copy DB, migrate, and test end to end

See how to use dump done when migrating from centos 6 to cents 7.

PR #724, opened 2016-05-24
Issue #694 migration es

st:
connected to https://github.com/slipstream/SlipStreamServer/issues/694


ssclj/jar/src/main/clojure/com/sixsq/slipstream/ssclj/middleware/cimi_params_impl.clj
```diff
@@ -115,12 +115,14 @@
...
   [{:keys [params] :or {:params {}} :as req}]
   (if-let [filter-param (get params "$filter")]
-    (->> filter-param
-         (as-vector)
-         (wrap-join-with-and)
-         (parser/parse-cimi-filter)
-         (throw-illegal-for-invalid-filter)
-         (add-cimi-param req :filter))
+
+    (add-cimi-param req :filter filter-param)
```
loomis: This change seems to be causing all of the tests in `cimi_params_impl_test.clj` using `process-filter` to fail.  Before this was returning the parsed filter, now this is returning the request with the raw, unparsed filter.  Which should be the correct behavior?

PR #727, opened 2016-05-27
parameter for /auth/login can be username or user-name

st:
Note that if both `user-name` and `username` are provided, 401 is
returned.

@rbf check especially auth/src/com/sixsq/slipstream/auth/internal.clj

Connected to https://github.com/slipstream/SlipStreamDocumentationAPI/issues/28


auth/src/com/sixsq/slipstream/auth/internal.clj
```diff
@@ -10,13 +10,20 @@
...
 
+(defn- xor
+  [a b]
+  (and (not (and a b))
+       (or a b)))
+
 (defn- extract-credentials
   [request]
-  (uh/select-in-params request [:user-name :password]))
+  (let [username (->> request :params ((juxt :user-name :username)) (apply xor))]
```
rbf: Congrats for that one. 👏😳

Do we need to support the legacy `user-name` param?
If yes, do we really want to cancel the value if both are present? Is it really worth the complexity?
If yes, is there a test for a request with both params `username` and `user-name`?

If we only want to give preference to the new param `username` with the legacy `user-name` as a fallback, following code would be probably simpler:

```clojure
(->> request :params ((some-fn :username :user-name)))
```

st: Supporting `user-name` seems reasonable to me.
Yes, there is a test checking inconsistent parameters:
https://github.com/slipstream/SlipStreamServer/pull/727/files/3e832a3edcfb84e5456687a483dc169efa23b599#diff-18000da1b1f36b867e0a4cd103ae1df3R22

rbf: Ok, thanks for the clarification. Merging.

loomis: I think that the best behavior is to prefer "username" and fallback to "user-name".  If both are provided, only the "username" value is used.  I think changing the order of the keywords and switching `xor` to `or` accomplishes this.

rbf: I just merged before seeing your comment, sorry.
The behaviour you mention is accomplished by the `some-fn` example.

Issue #735, opened 2016-06-01
create /ui/placement resource


konstan:
- needs authn: use auth as impl. example (not CIMI resource)
- contact PRS micro-service
- initial implementation 
  - empty in -> clouds out

konstan:
Depends on #736

st:
- [x] new service in ssclj : /ui/placement (echo service)
- [x] java proxy to this new service
- [ ] contact PRS-lib
- [ ] JS calls /ui/placement on local server (no cross site calls)

st:
JS code is manually tested in a developer console:

```javascript
SlipStream.request
  .put("/ui/placement")
  .data({clouds:["aa", "bb"]})
  .serialization("json")
  .dataType("json")
  .onSuccess(function(x){console.log(x["clouds-mockup-1"]);})
  .send()
```

With this call, we see that the mockup `/ui/placement` is used.

PR #749, opened 2016-06-10
Issue #735 create ui placement

st:
Connected to #735


jar-service/src/main/java/com/sixsq/slipstream/ui/UIPlacementResource.java
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ *
+ */
+public class UIPlacementResource extends BaseResource {
+
+    private static Logger logger = Logger.getLogger(UIPlacementResource.class.getName());
+
+    @Get("json")
+    public Representation getUI() {
```
konstan: What is the purpose of the `GET`?  
IMHO, if the operation doesn't logically apply to the facade, then I wouldn't even implement the stub.

st: It is convenient for quick checking from the browser that the service exists.

konstan: What "service" do you mean here?  Is it PRS?  Are you planning to call PRS directly from within the `GET`?  Because `/ui/placement` is a resource on SS service (SSCLJ later) and the current `GET` call is just a check for the resource availability, which I guess will not be used on production.  Do we need it then?

st: No. this was a check for `ui/placement` availability. Will remove this.

Issue #732, opened 2016-06-01
Configuration parameter in '/configuration' for PRS endpoint


rbf:
It will be used by the client (javascript) to update the UI.

PR #759, opened 2016-06-14
Issue #735 create ui placement

st:
Connected to #735

### DIFF
--- a/jar-prslib/pom.xml
+++ b/jar-prslib/pom.xml
@@ -26,6 +26,11 @@
       <artifactId>default-deps</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>com.sixsq.slipstream</groupId>
+      <artifactId>SlipStreamClientAPI-jar</artifactId>
+      <version>${project.version}</version>
+    </dependency>
   </dependencies>
   
   <profiles>
--- a/jar-prslib/pom.xml
+++ b/jar-prslib/pom.xml
@@ -26,11 +26,6 @@
       <artifactId>default-deps</artifactId>
       <version>${project.version}</version>
     </dependency>
-    <dependency>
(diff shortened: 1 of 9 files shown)


konstan:
Related to https://github.com/slipstream/SlipStreamServer/issues/732

jar-prslib/pom.xml
```diff
@@ -26,6 +26,11 @@
       <artifactId>default-deps</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>com.sixsq.slipstream</groupId>
+      <artifactId>SlipStreamClientAPI-jar</artifactId>
+      <version>${project.version}</version>
+    </dependency>
```
konstan: This is not needed as this dependency is expressed already in build.boot.

jar-service/src/main/java/com/sixsq/slipstream/run/PlacementValidator.java
```diff
@@ -14,7 +17,23 @@
 
     private static Logger logger = Logger.getLogger(PlacementValidator.class.getName());
 
+    private static boolean isPlacementEnabled = false;
+    static {
+        try {
+            isPlacementEnabled = Configuration.isEnabled(UIPlacementResource.PRS_ENABLED_PROPERTY_KEY);
+            logger.info("Placement Server enabled ? " + isPlacementEnabled);
```
konstan: Lets log on `fine` level.

st: Even if logged only once per application startup?

konstan: Agreed - `info`.  But it should go outside of the try/catch.

Issue #771, opened 2016-06-22
[PRSlib] artefact edition is hardcoded in the build.boot


konstan:
Instead, it can be provided at deploy time in pom.xml with 

`boot --set-env edition=${slipstream.edition} mvn-deploy`

PR #772, opened 2016-06-22
don't define edition in build.boot and provide it from pom.xml

konstan:
Connected to #771

### DIFF
--- a/jar-prslib/build.boot
+++ b/jar-prslib/build.boot
@@ -5,14 +5,13 @@
         repo-type (if (re-find #"SNAPSHOT" version)
                     "snapshots"
                     "releases")]
-    (str nexus-url repo-type "-" edition "-rhel7")))
+    (str nexus-url repo-type edition "-rhel7")))
 
 (set-env!
   :jar-name "SlipStreamServerPRSlib-jar"
   :project 'com.sixsq.slipstream/SlipStreamServerPRSlib-jar
   :version +version+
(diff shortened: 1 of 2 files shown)


konstan:
@loomis If the change is approved and merged, the same change should be done in https://github.com/slipstream/SlipStreamClientAPI/tree/master/jar

loomis:
But by making this change, it means that every time `boot` is used from the command line, the correct value for the edition will have to be set explicitly to pull in the right clojure client API.  So every `boot test` will become `boot test --set-env edition=community`, etc.  I'm not convinced this is an improvement.

konstan:
Then, we can to define `edition` in `sixsq/boot-deputil` (or similar).  OK?

loomis:
The task for managing dependencies isn't well-suited to handling other parts of the global configuration.  Ideally you'd like to define a function that could be used to set the value for "edition".  Something like this could be achieve with the `boot.profile`, but the code would then need to be duplicated to every repository.  I'm wondering whether this really needs to be solved through automation/global configuration.  There are only two values for this and once set correctly in the various repositories, git does a good job of not overwriting the value.

loomis:
The build file must always reference some source file that contains the correct configuration for the edition.  In this case, either we have to hardcode that reference or hardcode the name of the edition.  For the dependencies, there is a gain by using the reference because there are a large number of values.  For the edition, it don't see the gain as this is a single value.  I'd propose closing this PR as it doesn't (IMO) bring any real benefit.

konstan:
Agreed and closing.

Issue #773, opened 2016-06-22
[PRSlib] hardcoded SS dependencies


konstan:
They have to be managed with `sixsq.boot-deputil/set-deps!`

PR #774, opened 2016-06-22
manage SS dependencies using sixsq.boot-deputil/set-deps!

konstan:
Connected #773


loomis:
Please move all of the non-boot-task-related dependencies to the deps.edn file.

konstan:
Done in df02386

PR #794, opened 2016-08-05
remove dependency on hsqldb

loomis:
Connected to slipstream/SlipStreamServerDeps#47.


loomis:
References to the ssclj database in hsqldb have been removed where possible.  The production configuration file still needs to reference this because the migration scripts depend on this.

Issue #793, opened 2016-08-04
Improve collector


st:
On nuv.la, the dashboard experienced an issue with the display of the VMs.
Restarting nuv.la fixed the issue.
This lead us to investigate the Collector. (jar-async project).

#### Report of an excessive usage of the collector channel

The size if the channel is 512. We log the current usage (by counting the number of elements of the set `@busy`).
The atom `@busy` observes the number of tuples (user, connector) being processed.
Theoretically, the count of `@busy` can not go beyond `nb-users` * `nb-connectors`, because `@busy` is a set. A gross approximation is 9'000. (300 * 30). 
How can we end up with more than 24000 elements?

Secondly, there are protections to prevent insertions of requests in channel if busy. They did not work. 

```
2016-08-04T11:07:10.726+0000 INFO com.sixsq.slipstream.util.Logger info all users: 24819 / 512 => 4847.4609375 %
 2016-08-04T11:07:17.088+0000 INFO com.sixsq.slipstream.util.Logger info online users: 24820 / 512 => 4847.65625 %
 2016-08-04T11:07:27.378+0000 INFO com.sixsq.slipstream.util.Logger info online users: 24819 / 512 => 4847.4609375 %
```

#### Spikes of usage

On nuv.la, with minimal activity (2 users), we see regular spikes of channel usage. In the following logs, the `78` indicates that during the 10 seconds loop for active users, the collect was not completed.

```
 2016-08-04T16:09:50.485+0000 INFO com.sixsq.slipstream.util.Logger info online users: 3 / 512 => 0.5859375 %
 2016-08-04T16:10:00.815+0000 INFO com.sixsq.slipstream.util.Logger info online users: 78 / 512 => 15.234375 %
```

Command to display the 20 largest usage of channel by extracting the logs:
```
egrep ' => ' /opt/slipstream/server/logs/slipstream.log.0 | awk '{ print $11 }' | sort -nr | head -20
```

#### Insert requests for configured connectors 

Java code should return the tuple themselves, and we shouldn't do naively the scalar product. Expected result : [ [u1, [c1, c2]], [u2, [c3]].

#### Change logic of marking a tuple busy

Around this line
https://github.com/slipstream/SlipStreamServer/blob/master/jar-async/src/slipstream/async/collector.clj#L199
the marking is done just before inserting the collect request. We could mark this tuple not after having wait the spacing delay.

#### More granular spacing of requests

[...]

PR #798, opened 2016-08-09
Issues #793 improve collector

st:
Connected to #793

### DIFF
--- a/rpm/src/main/systemd/slipstream.service
+++ b/rpm/src/main/systemd/slipstream.service
@@ -1,7 +1,7 @@
 [Unit]
 Description=SlipStream Service
 After=syslog.target network.target
-After=hsqldb.service
+After=hsqldb.service ssclj.service
 
 [Service]
 RemainAfterExit=yes
--- a/jar-async/build.boot
+++ b/jar-async/build.boot
@@ -1,0 +1,50 @@
+(def +version+ "3.10-SNAPSHOT")
+
+(set-env!
+  :project 'com.sixsq.slipstream/SlipStreamAsync
+  :version +version+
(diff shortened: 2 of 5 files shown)


rpm/src/main/systemd/slipstream.service
```diff
@@ -1,7 +1,7 @@
 [Unit]
 Description=SlipStream Service
 After=syslog.target network.target
-After=hsqldb.service ssclj.service
+After=hsqldb.service
```
konstan: Are you sure this change should be part of this PR?  Also, I guess by doing so you are rolling back changes from Cal from yesterday.

loomis: I've put ssclj.service back into the service file.

ssclj/rpm/src/main/systemd/ssclj.service
```diff
@@ -1,7 +1,8 @@
 [Unit]
 Description=SlipStream API
-After=syslog.target network.target
-After=hsqldb.service elasticsearch.service
+After=syslog.target
+After=network.target
+After=hsqldb.service
```
konstan: @st If I recall correctly, hsqldb is no longer needed by ssclj.  I think, the `After` should read like this

```
-After=syslog.target network.target
 -After=elasticsearch.service
```
Please check.

loomis: No, hsqldb is still needed because of the authentication.  The checking/adding of users is done directly in the hsqldb rather than through the SlipStream interface.  Once the user resource moves to clojure, this dependency can be removed.

Issue #810, opened 2016-08-15
mark ssclj exit code 143 as successful exit


loomis:
The JVM returns the exit code 143 when being killed via SIGTERM.  This is the normal shutdown mode from systemd.  Mark this as a normal exit in the systemd unit file.

PR #811, opened 2016-08-15
allow 143 as successful exit

loomis:
Connected to #810.


ssclj/rpm/src/main/systemd/ssclj.service
```diff
@@ -19,5 +19,10 @@ ExecStart=/usr/bin/java \
             com.sixsq.slipstream.ssclj.app.main $SSCLJ_PORT
 ExecStop=/bin/kill -TERM $MAINPID
 
+./ssclj/rpm/src/main/systemd/ssclj.service
```
konstan: What is this for on line 22? Looks like copy/paste issue?

loomis: Indeed.  Removed.

Issue #779, opened 2016-07-04
migrate auth build from maven to boot


loomis:
Migrate the build of the auth module from maven to boot.

loomis:
Trying to do this has revealed a problem with dependency management.  When building all of the modules in one go, maven will optimize the build and use the dependencies directly from the `pom.xml` file rather than reading the generated `pom.xml` file.  Currently, the full dependency list is only included in the boot build, so only the generated `pom.xml` file has the complete list.  Because of this optimization, the full build will fail, while an individual build of the module will succeed.  Possible solutions to this problem are:

 * Provide complete dependencies in both the maven and boot files.
 * Use the (resolved) maven dependencies from boot (see [maven interop task](https://github.com/andrewmed/boot-maven-interop)).
 * Move those repositories that depend on other modules to boot first (and assume that boot will handle this case correctly).

The first option would require dependency management in two places and isn't desirable.  The referenced maven interop task, only reads the local `pom.xml` file, so it won't work with our central management of dependency versions.  It seems that we're left only with the last option, checking that it indeed does work correctly.

PR #814, opened 2016-08-15
use boot to build auth module

loomis:
Change to `boot` for building of auth module.  To have the dependencies picked up correctly in `maven`, there a separate dependencies for `com.sixsq.slipstream/auth` and `com.sixsq.slipstream/auth-dep`.  The second one just handles the submodule build dependencies, the first is the actual artifact.  By not directly including the artifact in the submodule build, `maven` will read the associated pom file and dependencies from those generated by `boot`. 

Connected to #779.


konstan:
Just a comment. Had to pull and build `SlipStream` (that defines version for `com.sixsq.slipstream/auth-de`) to be able to build `SlipStreamServer`.
Checking further.

loomis:
Yes, forgot to say something about that.  I pushed that change (additional dependency) directly to master as this is backwards compatible.

Issue #840, opened 2016-08-28
improve lifecycle test functions


loomis:
The lifecycle test functions perform common `is` tests on CIMI resources. Unfortunately when they fail, it is difficult to trace where the failure was.  This is because the `is` macro takes the source information from the surrounding function. 

Do the following improvements:

 - [ ] Switch all `is`-containing functions to macros.
 - [ ] Rename the `body->json` function to `body->edn`, which better reflects what it does.
 - [ ] Provide messages for all tests to make it clear what is failing.

PR #841, opened 2016-08-28
improve the lifecycle test functions

loomis:
Connected to #840.


ssclj/jar/src/com/sixsq/slipstream/ssclj/resources/common/utils.clj
```diff
@@ -23,7 +23,9 @@
 ;; NOTE: this cannot be replaced with s/lisp-case because it
 ;; will treat a '/' in a resource name as a word separator.
 (defn de-camelcase [str]
```
loomis: Function will now gracefully handle nil, rather than throwing a NPE.

Issue #676, opened 2016-03-28
add Session and SessionTemplate resources


loomis:
Add these resources to allow the URLs for logging into/out of the server to be auto-discovered.

loomis:
Things to be done to finalize the code:

 - [ ] Bring the code up to date with the current master.
 - [ ] Implement the SessionTemplates for all of the supported access methods.
 - [ ] Follow the 'describe' pattern used with ConnectorTemplates to allow the UI to discover the required fields.
 - [ ] Find a way for the clojure/script client to determine the session being used by the client.

loomis:
The other authentication methods will be added once we have feedback on the session templates for the internal authentication.  The implementation may need some significant refactoring to make it flexible enough to support all of the authentication methods with configuration.

PR #842, opened 2016-08-29
add prototype session resources

loomis:
Connected to #676.

Move the login/logout functionality to CIMI-like resources.  Would allow for automatic discovery of login methods and make the system more extensible.


loomis:
Do not merge this PR yet; it is just for review.  The implementation needs to avoid the copying of cookie-related utilities and implementations for GitHub and CYCLONE login methods need to be added.

loomis:
This branch is no longer current.  Use #973 instead.

Issue #800, opened 2016-08-12
providing invalid filter responds with internal server error


loomis:
Providing an invalid filter as a query parameter for a search request (e.g. on https://nuv.la/api/event), causes the server to respond with an internal server error.

The error should instead be a 400 (client error) and the error message should indicate clearly what the problem was.

st:
- [x] change middleware order and add specific exception for bad CIMI param
- [ ] add unit tests

PR #851, opened 2016-09-21
Issue #800 invalid filter wrong exception

st:
Connected to #800

### DIFF
--- a/ssclj/jar/src/com/sixsq/slipstream/ssclj/middleware/cimi_params_impl.clj
+++ b/ssclj/jar/src/com/sixsq/slipstream/ssclj/middleware/cimi_params_impl.clj
@@ -9,7 +9,8 @@
     [com.sixsq.slipstream.ssclj.middleware.accepted-mime-types :as mime]
     [superstring.core :as s]
     [instaparse.core :as insta]
-    [com.sixsq.slipstream.ssclj.resources.common.debug-utils :as du]))
+    [com.sixsq.slipstream.ssclj.resources.common.debug-utils :as du]
+    [com.sixsq.slipstream.ssclj.resources.common.utils :as u]))
 
 
 (def ^:const default-last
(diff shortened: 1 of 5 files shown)


ssclj/jar/src/com/sixsq/slipstream/ssclj/resources/common/utils.clj
```diff
@@ -72,6 +77,12 @@
     (str "undefined action (" (name request-method) ", " action ") for " uri)
     404 uri))
 
+(defn ex-bad-CIMI-filter
+  [filter-param]
+  (-> (str "Wrong CIMI filter [" filter-param "], "
```
loomis: "Invalid CIMI filter" would be better than "Wrong CIMI filter".

ssclj/jar/src/com/sixsq/slipstream/ssclj/middleware/cimi_params_impl.clj
```diff
@@ -103,9 +104,9 @@
   (s/join " and " (map #(str "(" % ")") filters)))
 
 (defn throw-illegal-for-invalid-filter
-  [parse-result]
+  [filter-param parse-result]
   (if (insta/failure? parse-result)
-    (throw (Exception. (str "Invalid filter: " (insta/get-failure parse-result))))
+    (throw (u/ex-bad-CIMI-filter filter-param))
```
loomis: Don't you lose the actual parsing error by ignoring `parse-result`?  Having that error message is likely to be more useful than citing the standards document.

ssclj/jar/src/com/sixsq/slipstream/ssclj/resources/common/utils.clj
```diff
@@ -72,6 +77,12 @@
...
+           "Please consult 4.1.6.1 Filtering Collections of https://www.dmtf.org/sites/default/files/standards/documents/DSP0263_2.0.0.pdf")
```
loomis: See above.  I don't think citing the standard is useful.

Issue #785, opened 2016-07-18
Instance ID is not filled by SS client when error occurs during deployment


0xbase12:
This [run](https://nuv.la/run/19f78151-3f0a-4894-a5f8-896ae809fa5f) have failed due to a lack of ip  (cf. run-cancelled.png) and is indicated as "Cancelled". It not displayed in my active deployment (no-deploy), but the VM is still running on LAL cloud (vm-still-active). On the VM tab no vm can be killed (no-vm-killable). By elimination i was able to indicate which vm is the one we are talking about, but no button to kill it.
I can kill the vm if i click on Terminate (cf. run-cancelled.png) but i would never have re-openened this page it if i didn't saw that the vm was still alive on lal cloud web interface.
This [run](https://nuv.la/run/fb3d1384-705f-4c55-a243-a9b17a8f310e) is in the same state.

Could you please find a way to keep the deploys visible in Deployments as long as there is VM running in order to let user kill these zombies ?
![no-deploy](https://cloud.githubusercontent.com/assets/5518233/16909592/0847418e-4cd4-11e6-9b8a-585be2e5badb.png)
![no-vm-killable](https://cloud.githubusercontent.com/assets/5518233/16909590/08449bfa-4cd4-11e6-99eb-bb7b8fec1eba.png)
![run-cancelled](https://cloud.githubusercontent.com/assets/5518233/16909591/0846c36c-4cd4-11e6-88ed-0015c59fcb9d.png)
![vm-still-active](https://cloud.githubusercontent.com/assets/5518233/16909593/084d52fe-4cd4-11e6-9314-665b56371a44.png)

PR #852, opened 2016-09-21
Rewrite of static the method parseRunInstanceResult ...

schaubl:
...to be more smart in the retrieval of instance id and ip.
Connected to #785

### DIFF
--- a/jar-connector/src/test/java/com/sixsq/slipstream/connector/CliConnectorBaseTest.java
+++ b/jar-connector/src/test/java/com/sixsq/slipstream/connector/CliConnectorBaseTest.java
@@ -1,0 +1,94 @@
+package com.sixsq.slipstream.connector;
+
+/*
+ * +=================================================================+
+ * SlipStream Server (WAR)
+ * =====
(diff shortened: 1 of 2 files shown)


jar-connector/src/test/java/com/sixsq/slipstream/connector/CliConnectorBaseTest.java
```diff
@@ -0,0 +1,94 @@
...
+            if (output.contains(ip)) assertEquals(message, ip, resIp);
+        }
+
+        ArrayList<String> outputsWithException = new ArrayList<>();
+        outputsWithException.add(",");
+        outputsWithException.add("");
+        outputsWithException.add(","+ip);
+
+        for (String output: outputsWithException) {
+            Logger.info("Output: " + output);
```
konstan: Is this logging needed here?

schaubl: No, you're right, I did this while debugging this unit test but now it's part of the error message if the assertion fail. I will remove it.

Issue #817, opened 2016-08-16
store/load ServiceConfiguration entity to/from Elasticsearch as json


konstan:

konstan:
Demoed from branches https://github.com/slipstream/SlipStreamServer/tree/issue_%23817-serviceconfig-to-es
and 
https://github.com/slipstream/SlipStreamConnectors/tree/feature/service-config-in-es

on Fri Sep 16 CEST 2016

konstan:
TODO

- [ ] add editable `description` parameter for the connector instance itself with the default `Instance of connector {{connector name}}`,
- [ ] migrate other connectors (build list of current connectors and vote for the most important ones (probably the ones that are now used on nuv.la)),
- [ ] finalise `ss-config` utility that should load documents to DB (ES), use edn format for the input,
- [ ] updated SlipStream deployment regarding the new way of the service configuration.

konstan:
Here is the list of the first candidate connectors for migration to ssclj service.  Those are the ones currently used on nuv.la

- [ ] azure
- [ ] cloudstack
- [ ] ec2
- [ ] exoscale
- [ ] nativesoftlayer
- [ ] nuvlabox
- [ ] opennebula
- [x] openstack (done in https://github.com/slipstream/SlipStreamConnectors/tree/feature/service-config-in-es)
- [ ] stratuslabiter

loomis:
My personal opinion is that azure and nativesoftlayer could be deferred, if necessary.

mebster:
Yes.  But then we need to remove these connectors from Nuvla.

mebster:
An important issue raised and briefly discussed during the demo is support for i18n.  I've created an new issue to track that: https://github.com/SixSq/tasklist/issues/862.

konstan:
Demoed on Fri Oct  7 2016.
- building of enterprise SS from the feature branches
- deployment of 3.13 from enterprise snapshots
- migration of configuration from nuv.la to the deployed instance
- new ss-conf* utilities for the service configuration migration and update of the service configuration.

TODO

- [ ] documentation of migration procedure (to go to release notes)
- [ ] documentation of the new server configuration method and used utilities with examples.

PR #857, opened 2016-10-10
Service Configuration in Elasticsearch

konstan:
This set of changes moves the persistence of ServiceConfiguration entity from hsqldb to Elasticsearch as a set of two types of the documents
- service configuration itself (`configuration/slipstream` resource in SSCLJ)
- per connector instance configurations (`connector/<connector instance name>` resources in SSCLJ).

It also brings utility `ss-config` for managing the configuration via Elasticsearch and the migration utility `ss-config-migrate`.

I think it's better to squash when merging.

Here are the corresponding PR for connectors
https://github.com/slipstream/SlipStreamConnectors/pull/122
https://github.com/SixSq/SlipStreamConnectors/pull/33
https://github.com/SixSq/SlipStreamConnector-EC2/pull/27
https://github.com/SixSq/SlipStreamConnector-NuvlaBox/pull/9

Connected to #817


schaubl:
I've added @st to this PR so he can review the Clojure part and I will review the Java part.

Issue #877, opened 2016-11-14
refactor micro-service


loomis:
We are starting to have a number of micro-services in clojure that reuse the same basic ring application container. Copies of the code exist in all of the micro-services.

Refactor this to have a single application server package that can be required by the various micro-services, which then simply provide their own "defaults" and systemd service files.

loomis:
Refactor the PRS to use the generic ring container.  Before doing so, the code needs to be cleaned up (see #884).

PR #883, opened 2016-11-20
Feature/refactor webapp

loomis:
Connected to #877. 

Please review this PR, but do not merge it.  At least one of the real services should be refactored to use this generic service before merging the changes.


ring-container/src/sixsq/slipstream/server/ring_container.clj
```diff
@@ -0,0 +1,177 @@
...
+                 (log/error "failure when shutting down ring container:" (str e))))))))
+
+(defn- start-container
+  "Starts the aleph container with the given ring handler and on the given
+   port. Returns the server object that has been started, which can be stopped
+   by calling 'close' on the object."
+  [handler ^long port]
+  (log/info "starting aleph application container on port" port)
+  (let [start-server (dyn-resolve 'aleph.http/start-server)]
+    (log/info "started aleph application container on port" port)
```
konstan: This log message should go after `(start-server handler ...)`.

ring-container/src/sixsq/slipstream/server/ring_container.clj
```diff
@@ -0,0 +1,177 @@
...
+               (catch Exception e
+                 (log/error "failure when shutting down ring container:" (str e))))))))
+
+(defn- start-container
+  "Starts the aleph container with the given ring handler and on the given
+   port. Returns the server object that has been started, which can be stopped
+   by calling 'close' on the object."
+  [handler ^long port]
+  (log/info "starting aleph application container on port" port)
+  (let [start-server (dyn-resolve 'aleph.http/start-server)]
```
konstan: Why aleph is dynamically resolved instead of required explicitly?

ring-container/src/sixsq/slipstream/server/ring_container.clj
loomis: Yep.  Fixed in latest commit.

ring-container/src/sixsq/slipstream/server/ring_container.clj
loomis: When started as a daemon through the `main` function, the java class loader will be used.  As this class loader knows nothing about clojure source files, everything it needs to load the class must be available as compiled class files.  If aleph, environ, etc. were statically required, then the AOT compilation would also need to include those libraries as well, bloating the size of the jar file and causing potential conflicts with those dependencies on the classpath.

The AOT compilation of the`ring-container` namespace could be further reduced by removing the logging or also requiring it dynamically.  IMO, that seemed to be too much work for too little gain, but other opinions are welcome.

loomis:
@konstan Please review again and merge if OK.  We'll use the PRS as the guinea pig for the change, seeing as we're having problems with it anyway.

Issue #874, opened 2016-11-07
provide json representation of modules


loomis:
Provide a rudimentary JSON representation of modules to allow chooser and searching facilities through the ClojureScript API.

loomis:
A simpler solution may be to just do an automated translation of XML to JSON, rather than trying to generate the JSON from the underlying objects.  See the [documentation](http://javadox.com/org.json/json/20151123/org/json/XML.html) for the org.json library.

PR #890, opened 2016-12-06
add simple conversion of xml to json

loomis:
Connected to #874.

### DIFF
--- a/jar-service/pom.xml
+++ b/jar-service/pom.xml
@@ -158,6 +158,11 @@
 		</dependency>
 
 		<dependency>
+			<groupId>org.json</groupId>
+			<artifactId>json</artifactId>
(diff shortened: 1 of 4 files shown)


loomis:
There is an open issue that needs more investigation: any client that doesn't explicitly specify a media type may get either XML or JSON.  This may cause clients that were expecting XML by default to fail.  We need to determine if all of our clients send requests with explicit media type(s).

loomis:
Looking through the client code briefly, it appears that we always explicitly ask for XML when we need it. So, in principle providing JSON in addition to XML should not break anything.  I've checked that asking for JSON does return a correct JSON document with a literal translation of the XML content.  This works for all module types and for the run resource.

Issue #685, opened 2016-04-06
Tags order modified


WinstonSureChill:
When deploying an application, I specify tags, and it appears that sometimes Slipstream doesn't store the tags in the same order I initially specified them.

For example, when I deploy an application with tags `elouan, dev, DNS-namespaces`, tags are stored in this different order: `DNS-namespaces, dev, elouan`.

However, when I deploy a single component (not a whole application) with the same tags (`elouan, dev, DNS-namespaces`), the order is preserved.

For a little bit of context, my team is planning to do some introspection using tags, and in our code we expect them to be in the specific order in which we specified them initially.

As a solution, we can get rid of the order by using the following tag format: `key1_value1, key2_value2, key3_value3`, but it makes it heavier and we would prefer to rely on the order.

WinstonSureChill:
Finally we decided not to rely on the tag order

mebster:
Hi.  Thanks for sharing this.  We'll discuss this issue and either make it clear in the documentation that tags are unordered, or make it ordered in all cases.

rbf:
For reference, the server is not involved on that: the client sanitises the tag list before sending it to the server by removing duplicates and sorting it alphabetically. As correctly reported by @WinstonSureChill, the deploy dialog in the component page does not include the  sanitation code.

I've opened an issue in the UI module for this: https://github.com/slipstream/SlipStreamUI/issues/577.

PR #891, opened 2016-12-07
Issue #685 instance type at runtime

st:
Connected to https://github.com/slipstream/SlipStreamUI/issues/685
To be merged with other PRs (UI and Openstack connector).


loomis:
The change for the UI must be reapplied.  See slipstream/SlipStreamUI#700 and the associated PR for what needs to be uncommented in `modal_dialogs.js`.

PR #906, opened 2017-01-17
publish prs-service-dep

loomis:
Connected to SixSq/tasklist#977.

This turns out only to be a problem for the PRS Service jar file.  All the other "dep" repositories already install/deploy the pom files.


loomis:
Make a mess of this.  Deleting...

Issue #905, opened 2017-01-17
remove service catalog submodules


loomis:
The service catalog submodules that create a separate jar and rpm are a hold over from when the service catalog was an enterprise-only feature. Merge the service catalog resources into the main cimi server (ssclj) and remove the separate rpm.

PR #908, opened 2017-01-17
merge service catalog into main server

loomis:
Connected to #905.


konstan:
We need to mark `slipstream-service-catalog` RPM as obsolete or conflicting in `slipstream-ssclj` RPM, otherwise `slipstream-service-catalog` will not be removed during the next upgrade.

konstan:
@loomis please check previous comment and 0e08ae1

loomis:
@konstan Change looks good to me.  We merge?

Issue #912, opened 2017-02-01
move nuvlabox admin ui


loomis:
Move the NuvlaBox administration UI code from the enterprise edition to the community edition. Server aspects remain with the enterprise edition.

PR #913, opened 2017-02-01
move nuvlabox admin ui to community

loomis:
Connected to #912


loomis:
The changes have been mixed between this change and those for SixSq/tasklist#983.

Issue #919, opened 2017-02-06
debug unit test failures on centos 7


loomis:
The build job for SlipStreamServer has failing unit tests on CentOS 7. The tests do not fail on Mac OS.

konstan:
Just noticed this.  This is due to a different order of the tests.  And yes, the ES DB bootstrap is needed.  There is this helper method to solve the issue.

```
import com.sixsq.slipstream.es.CljElasticsearchHelper;
...
        @BeforeClass
	public static void setupClass() throws ConfigurationException, SlipStreamException {
		CljElasticsearchHelper.createAndInitTestDb();
       ...
       }
```

PR #921, opened 2017-02-07
add configuration initialization

loomis:
Connected to #919


konstan:
Just noticed this.  This is due to a different order of the tests.  And yes, the ES DB bootstrap is needed.  There is this helper method to solve the issue.

```
import com.sixsq.slipstream.es.CljElasticsearchHelper;
...
        @BeforeClass
	public static void setupClass() throws ConfigurationException, SlipStreamException {
		CljElasticsearchHelper.createAndInitTestDb();
       ...
       }
```

Issue #939, opened 2017-02-28
default elasticsearch parameter values


loomis:
If the values for the Elasticsearch host and port are not provided, use localhost and the default port. This simplifies the configuration of the services when testing.

PR #944, opened 2017-03-01
use host localhost and port 9300 by default

0xbase12:
Connected to #939


loomis:
The change for the default parameters is fine.

In the spirit of "leave the place cleaner than you found it", I've cleaned up the logging, removed some unnecessary private functions, and fixed some things picked up by the static analyzers.  @0xbase12  @konstan check that the changes are OK.

Issue #941, opened 2017-02-28
provide resources for module/connector information


loomis:
Provide resources that provide the information that is currently collected for placement/pricing requests. These resources should provide module and user connector information.  Additional resources should also be provided if other information is required by the PRS.

PR #945, opened 2017-03-03
ui/placement resource

konstan:
- removed call to PRS
- simply return JSON built to call PRS
- updated unit tests
- removed placement validation for Run
- removed PRS from server configuration

Connected to #941


konstan:
Migration in ES DB will be required. The following config parameters should be removed from `configuration/slipstream`
```
:prsEndpoint "http://localhost:8203/filter-rank"
:prsEnable true
```

loomis:
The branch builds for me without problems. However, it seems that this requires changes on the UI and the PRS before this can be merged without breaking the deployments.  @konstan is this correct?

konstan:
Yes, this is correct.  Clients definitely should change before this can be merged.

Issue #948, opened 2017-03-06
[prs-service] add authentication check


konstan:
Only connection with a cookie containing a valid token should be allowed.

PR #952, opened 2017-03-06
added authn validation for PRS service.

konstan:
Connected to #948


loomis:
I made a few stylistic changes.  Otherwise the PR looks good. 

For information, I have in a separate branch a refactoring of the token module that pulls in the cookie manipulation code so that it doesn't need to be repeated everywhere.  I'll look into making that a separate PR.

Issue #949, opened 2017-03-06
make check of prefix/uri collisions more efficient


loomis:
The current implementation checks prefix/uri collisions by extracting all resources and then checking for collisions in the code.  Although the service isn't likely to have very large numbers of namespace resources, this is still not efficient. Change the implementation to use a query expression to select colliding resources at the database level.

PR #953, opened 2017-03-07
Feature/use db query

loomis:
Connected to #949.


loomis:
Wrong set of commits.  Moving to different PR.

Issue #961, opened 2017-03-15
refactor token module to provide cookie functions


loomis:
Refactor the token module so that it provides functions that create the required cookies directly, so that the code that implements the authentication methods is simpler.

0xbase12:
Is this ticket done? Can I close it?

PR #962, opened 2017-03-15
Feature/cookie functions

loomis:
Connected to #961.


jar-prs-service/src/sixsq/slipstream/prs/ring.clj
```diff
@@ -1,43 +1,19 @@
...
-      (log/warn (str "Error in extract-cookie-info: " (.getMessage ex)))
-      nil)))
-;; end.
+    [com.sixsq.slipstream.auth.cookies :as cookies]))
 
 (defn- authenticated?
   [request]
-  (not (nil? (seq (extract-cookie-info request)))))
+  (let [cookie (get-in request [:cookies "com.sixsq.slipstream.cookie"])]
+    (cookies/extract-cookie-info cookie)))
```
konstan: `authenticated?` is used in `if` for with `[]` is `true`.  I suggest, make a predicate out of it, i.e. returning only `true` or `false`.

PR #976, opened 2017-04-04
Fix performance issues with Module

schaubl:
Connected to #975 

- Use `FetchMode.SELECT` for `OneToMany` relations related to` Module` to prevent combinatory explosion with joins (cartesian product).
- Prevent SlipStream to load all `Modules` and all `Users` at startup to improve startup time.


schaubl:
FYI: These changes are already on nuv.la

Issue #1007, opened 2017-04-24
old cookie causes internal server error


loomis:
When accessing the v3.26 server with an old cookie that is still valid, the server responds with an internal server error (500).

PR #1008, opened 2017-04-24
attempted fix of 500 with old cookie

loomis:
Connected to #1007.


loomis:
I've deployed an updated SlipStreamService jar and the issue seems to be solved for me.

PR #1009, opened 2017-05-03
fix comparator for price ordering

loomis:
Connected to sixsq/SlipStreamPricing#41.

### DIFF
--- a/jar-placement-lib/src/sixsq/slipstream/placement/core.clj
+++ b/jar-placement-lib/src/sixsq/slipstream/placement/core.clj
@@ -137,13 +137,31 @@
        :disk          (disk service-offer)
        :instance_type (instance-type service-offer)})))
 
+(defn number-or-nil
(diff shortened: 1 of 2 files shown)


konstan:
It compiles and unit tests work.  As part of the validation, 

1. should I just push the jar with this change only to nuv.la and check if `java.lang.IllegalArgumentException: Comparison method violates its general contract!` is no longer thrown?  

2. Or I can create a system test for PRS with service catalog pre-populated with the service offers from nuv.la.  This will be an extension of nuv.la reference configuration that we have. That would actually work as a pre-prod test (on the data level).  Too heavy?  Do we need it now?

loomis:
I don't think that we need a system test for this at the moment.  Pushing it to Nuvla is reasonable.

konstan:
OK.  I'll validate this on nuv.la then.

konstan:
Tested on nuv.la.  Everything works.  Merging.

PR #1013, opened 2017-05-04
loosen event href schema; tighten other references

loomis:
Connected to #1012

### DIFF
--- a/ssclj/jar/src/com/sixsq/slipstream/ssclj/resources/spec/service_offer.cljc
+++ b/ssclj/jar/src/com/sixsq/slipstream/ssclj/resources/spec/service_offer.cljc
@@ -2,10 +2,14 @@
   (:require
     [clojure.spec.alpha :as s]
     [com.sixsq.slipstream.ssclj.util.spec :as su]
-    [com.sixsq.slipstream.ssclj.resources.spec.common :as c]))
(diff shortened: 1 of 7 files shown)


ssclj/jar/src/com/sixsq/slipstream/ssclj/resources/spec/service_offer.cljc
```diff
@@ -2,10 +2,14 @@
...
+    [com.sixsq.slipstream.ssclj.resources.spec.common :as c]
+    [com.sixsq.slipstream.ssclj.resources.spec.connector-template]))
 
-;; FIXME: This should use the standard href definition instead.
-(s/def :cimi.service-offer/href :cimi.core/nonblank-string)
+;;
+;; Must reference the raw name of the connector.  Reference the schema
+;; for defining connector names.
+;;
+(s/def :cimi.service-offer/href :cimi.connector-template.core/identifier)
```
0xbase12: Will clojure be able to get :cimi.connector-template.core/identifier with only a simple require?

loomis: Yes.  Spec maintains a central (global) registry of defined spec keywords.  The require will ensure that the definition in the other namespace is executed and then be available for this definition.

Issue #1018, opened 2017-05-08
session does not contain server name


loomis:
The session resource does not contain the server hostname.  This is because the server sets the header "slipstream-ssl-server-name" but the session code uses "slipstream-ssl-server-hostname". Fix the inconsistency.

PR #1019, opened 2017-05-08
fix creation of claims in cookies

loomis:
Connected to #1018


loomis:
While fixing this, I noticed that the claims were not correctly formatted.  They should be simple key/value pairs, but instead contained sub-structure (e.g. roles was a list).  This has also been corrected and deployed on Nuvla to validate.  This confirms that the full internal session lifecycle works correctly.

PR #1037, opened 2017-05-17
Feature/change session redirects

loomis:


ssclj/jar/src/com/sixsq/slipstream/ssclj/resources/session_template_github.clj
```diff
@@ -19,7 +19,15 @@
...
 ;; description
 ;;
-(def ^:const desc p/SessionTemplateDescription)
+(def ^:const desc
+  (merge p/SessionTemplateDescription
+         {:redirectURI {:displayName "Redirect URI"
+                        :category    "general"
+                        :description "Redirect URI"
+                        :type        "string"
+                        :mandatory   true
```
0xbase12: Should be False ?

loomis: Good catch! Changing...

PR #1053, opened 2017-05-29
Issue #1047 extend api manage cloudparameters at run creation

0xbase12:
Connected to #1047

### DIFF
--- a/jar-connector/src/main/java/com/sixsq/slipstream/factory/RunFactory.java
+++ b/jar-connector/src/main/java/com/sixsq/slipstream/factory/RunFactory.java
@@ -20,16 +20,10 @@ package com.sixsq.slipstream.factory;
  * -=================================================================-
  */
 
-import java.util.HashMap;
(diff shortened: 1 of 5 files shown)


jar-connector/src/main/java/com/sixsq/slipstream/factory/RunFactory.java
```diff
@@ -39,7 +33,8 @@
 
 public abstract class RunFactory {
 
-	public final Run createRun(Module module, User user) throws SlipStreamClientException {
+	public final Run
+	createRun(Module module, User user) throws SlipStreamClientException {
```
schaubl: Strange formatting. I would put `throws SlipStreamClientException {` in the second line instead of `createRun(Module module, User user)...`

0xbase12: 👍

PR #1064, opened 2017-06-02
Update pom.xml

0xbase12:
Connected to #1063


konstan:
Rejecting in favour of #1065 as the solution should be conditional.

Issue #988, opened 2017-04-18
cimi conventions for usage resources


loomis:
Refactor the usage resources to follow the CIMI conventions: attributes should be camel-cased, etc.

loomis:
Won't implement at this time.  See PR for rationale.

Issue #879, opened 2016-11-14
Rename REST resource usage to usage-summary


st:
- [x] `/api/usage ` -> `usage-summary`
- [x] `/usage` -> `/usage-summary` (URL in Java server is renamed but label stays unchanged)
- [x] in JSON responses `usages` -> `usage-summaries`
- [x] in JSON responses `usage` -> `usage-summary`
```
"usage-summaries" : [ {
   ...
    "usage-summary" : {
    ...
     },
    ...
  } ],
 ...
```
- [x] migration: rename existing documents
- [x] migration: improve : work by bulk, add option to script
- [x] use new resource in consolidation processes
- [x] use new resource in UI
- [ ] write migration documentation

A new resource `usage-summary` is created. A migration allows to copy existing `usage` documents in this new resource.
`usage` resource can then be deleted.

konstan:
@st Can this issue be closed?

st:
@konstan no, it was not merged yet.

schaubl:
Related branch: [issue_#879-rename-usage-summary](https://github.com/slipstream/SlipStreamServer/tree/issue_%23879-rename-usage-summary)

loomis:
Won't implement at this time.  See PR comments for rationale.

PR #1089, opened 2017-06-21
rename usage resources

loomis:
Connected to #879.
Connected to #988.


loomis:
This change was originally proposed to avoid a conflict with the dashboard and as a way of adding information about groups and roles to the usage resources, which will eventually be needed for HNSciCloud.  However, the changes require significant changes to the UI to work with the new names with little improvement to the UI or the API.  Moreover, the design of these resources does not follow the CIMI CRUD patterns and worse does not allow information like groups/roles and service offers to be added easily.  A wider discussion is needed to understand how to modify the usage monitoring system to permit use of richer information and a wider set of resources.  See SixSq/tasklist#1182.

Issue #1070, opened 2017-06-12
allow email addresses as usernames


loomis:
Change the regex for SlipStream usernames to allow valid email addresses. This is to allow the email to be used as a unique identifier (and SlipStream username) when logging in via an external identity provider.

loomis:
The [file/line](https://github.com/slipstream/SlipStreamServer/blob/master/jar-persistence/src/main/java/com/sixsq/slipstream/persistence/User.java#L389) with the java regex to be changed.

loomis:
Verify that the clojure-side schema will also allow usernames to be used in ACLs, etc.

loomis:
Put this on hold, as it isn't clear at the moment that this will be needed.

loomis:
When creating users programmatically the regex isn't checked.  Better to keep the restrictions on the UI to avoid potential conflicts with autogenerated names.  Nothing to do with this ticket.

loomis:
This is checked on updates to the user profile.  Either the regex needs to be changed or the check only applied when creating a new user.

PR #1102, opened 2017-07-04
only check username on new accounts

loomis:
Connected to #1070.


loomis:
The change has been deployed to Nuvla and appears to work.

PR #1107, opened 2017-07-06
threaded version, one request per connector implementation

0xbase12:
Connected to #1105 

fix placement params also:
use herited placement params at top level of place and rank of request and no more use of component placement param

### DIFF
--- a/jar-prslib/src/clj/sixsq/slipstream/prs/javawrapper.clj
+++ b/jar-prslib/src/clj/sixsq/slipstream/prs/javawrapper.clj
@@ -19,10 +19,9 @@
 (defn keyword-name
   "Function that converts a keyword to a string taking into account the keyword namespace."
   [k]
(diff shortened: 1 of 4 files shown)


jar-prslib/src/clj/sixsq/slipstream/prs/javawrapper.clj
```diff
@@ -15,6 +16,14 @@
     :name sixsq.slipstream.prs.core.JavaWrapper
     :methods [#^{:static true} [generatePrsRequest [java.util.Map] String]]))
 
+(defn keyword-name
+  "Function that converts a keyword to a string taking into account the keyword namespace."
+  [k]
+  (if (instance? clojure.lang.Keyword k)
```
schaubl:
```clojure
(if (keyword? k)
```

PR #1119, opened 2017-07-14
add implementation of more acl rights

loomis:
connected to SixSq/tasklist#1081


loomis:
This is backwards compatible with existing ACLs, so it can be merged without other changes to the authorization mechanisms.

Issue #1127, opened 2017-07-31
unable to deploy with v3.33 from UI


konstan:
Attempt to deploy from UI fails with 405

![screen shot 2017-07-31 at 11 07 28](https://user-images.githubusercontent.com/353767/28770737-80d23c2c-75e0-11e7-9f80-25d5ea649d91.png)

```
2017-07-31T09:05:08.181+0000 WARNING com.sixsq.slipstream.util.SscljProxy request ResourceException: Method Not Allowed
        with status: Method Not Allowed (405) - Method Not Allowed
        with content: {
  "status" : 405,
  "message" : "invalid method (get) for /api/null",
  "resource-id" : "/api/null"
}
```

It works when deploying with CLI.

konstan:
The issue is understood and is here https://github.com/slipstream/SlipStreamServer/blob/master/jar-persistence/src/main/java/com/sixsq/slipstream/util/ServiceOffersUtil.java#L49 where the `serviceOfferId` can be `null`.

After fixing this, there is another issue with

![screen shot 2017-07-31 at 12 10 35](https://user-images.githubusercontent.com/353767/28774691-f28566c4-75ee-11e7-9031-16137eeee2d7.png)

I'm assigning this to @elegoff 

The idea is to cover with unit-tests `ServiceOffersUtil`.

elegoff:
After second look analysis of this issue, it looks like a possible side effect about how to handle the Service Offer after https://github.com/slipstream/SlipStreamServer/pull/1122
Therefore reassigning to @schaubl

PR #1131, opened 2017-08-02
Unable to deploy from ui when no service offer provided

elegoff:
Connected to https://github.com/slipstream/SlipStreamServer/issues/1127


konstan:
This seems to be not fixed.  I deployed SS from enterprise snapshots and pushed locally compiled SlipStream*.jar's to the instance. However, after the slipstream service restart I'm still not able to deploy via Deploy dialog on UI

screen shot 2017-08-06 at 19 55 20

konstan:
After checking with @loomis - the above behaviour is correct.  I'm redeploying and testing again.

konstan:
The issue is fixed.  Merging.

PR #1139, opened 2017-08-07
Ensure module version atomicity

schaubl:
It doesn't ensure atomicity if SlipStream is scaled horizontally.
In the latter case the atomicity should be provided by the database.

Connected to #1135

### DIFF
--- a/jar-persistence/src/test/java/com/sixsq/slipstream/persistence/VersionCounterTest.java
+++ b/jar-persistence/src/test/java/com/sixsq/slipstream/persistence/VersionCounterTest.java
@@ -22,14 +22,53 @@ package com.sixsq.slipstream.persistence;
 
 import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertThat;
-
 import org.junit.Test;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.logging.Logger;
+
 public class VersionCounterTest {
+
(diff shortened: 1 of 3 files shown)


jar-persistence/src/test/java/com/sixsq/slipstream/persistence/VersionCounterTest.java
```diff
@@ -22,14 +22,53 @@
...
 		assertThat(VersionCounter.getNextVersion(), is(counter+1));
+		assertThat(VersionCounter.getNextVersion(), is(counter+2));
+	}
+
+	@Test
+	public void testParallelIncrement() throws InterruptedException {
+		List<Thread> threads = new ArrayList<>();
+
+		int initialVersion = VersionCounter.getNextVersion();
+		int finalVersion = initialVersion + numberOfThreads + 1;
```
konstan: Please set `finalVersion` as `final`, or move the sum directly to the assertion (line 63). In the latter case, define `initialVersion` as `final` as well.

schaubl: Why not having both as `final` in all cases ?

konstan: Yes, sure.

Issue #1174, opened 2017-08-29
Insert VirtualMachine documents in same code location where vm table is populated


elegoff:
In order to test ES documents insertion for VirtualMachine resource introduced by #1156 , we need a temporary solution to mimic what is done for hsqldb insertions (vm table)

PR #1178, opened 2017-08-30
Insert virtual machine resource

elegoff:
Connected to #1174


jar-connector/src/main/java/com/sixsq/slipstream/connector/Collector.java
```diff
@@ -358,7 +347,7 @@ private static void updateDbVmsWithCloudVms(VmsClassifier classifier, EntityMana
 			if (merge) {
 				logger.info("updateDbVmsWithCloudVms: Updating db VM: id=" + dbVm.getInstanceId() + ", state=" + dbVm.getState());
 				em.merge(dbVm);
-				VirtualMachineHandler.updateVM(dbVm);
+				VirtualMachineHandler.handleVM(dbVm);
```
schaubl: @elegoff I think you should extract this line out of the `if`. 
This will allow to copy all existing VMs to the new VirtualMachines resource.

Issue #1190, opened 2017-09-08
define cron process to insert global state


elegoff:
As part of https://github.com/SixSq/tasklist/issues/1302, we need a one minute process responsible for 
- Fetching the current global state 
- Inserting the documents (one by one until the bulk update API is available) into a dedicated resource (TODO : define resource name)

This inserted resource will be the basis of accounting calculations through aggregation queries

loomis:
The deployment of the metering service on the Elasticsearch node for Nuvla eventually caused the Elasticsearch database to crash. The error messages from the metering journal are similar to the following:
```
Exception in thread "async-dispatch-80" java.lang.IllegalStateException: org.apache.http.nio.reactor.IOReactorException: Failure opening selector
at org.apache.http.impl.nio.client.IOReactorUtils.create(IOReactorUtils.java:45)
at org.apache.http.impl.nio.client.HttpAsyncClientBuilder.build(HttpAsyncClientBuilder.java:666)
at org.elasticsearch.client.RestClientBuilder.createHttpClient(RestClientBuilder.java:202)
at org.elasticsearch.client.RestClientBuilder.build(RestClientBuilder.java:180)
at qbits.spandex.client_options$builder.invokeStatic(client_options.clj:197)
at qbits.spandex.client_options$builder.invoke(client_options.clj:192)
at qbits.spandex$client.invokeStatic(spandex.clj:83)
at qbits.spandex$client.invoke(spandex.clj:33)
at com.sixsq.slipstream.metering$fetch_global_state$fn__12058$state_machine__7262__auto____12061$fn__12063.invoke(metering.clj:55)
at com.sixsq.slipstream.metering$fetch_global_state$fn__12058$state_machine__7262__auto____12061.invoke(metering.clj:50)
at clojure.core.async.impl.ioc_macros$run_state_machine.invokeStatic(ioc_macros.clj:973)
at clojure.core.async.impl.ioc_macros$run_state_machine.invoke(ioc_macros.clj:972)
at clojure.core.async.impl.ioc_macros$run_state_machine_wrapped.invokeStatic(ioc_macros.clj:977)
at clojure.core.async.impl.ioc_macros$run_state_machine_wrapped.invoke(ioc_macros.clj:975)
at com.sixsq.slipstream.metering$fetch_global_state$fn__12058.invoke(metering.clj:50)
at clojure.lang.AFn.run(AFn.java:22)
[...]

PR #1197, opened 2017-09-20
Issue #1190 cron insert global state

elegoff:
Connected to https://github.com/slipstream/SlipStreamServer/issues/1190


global-state/jar/src/com/sixsq/slipstream/globalstate.clj
```diff
@@ -0,0 +1,83 @@
...
+
+
+(defn fetch-global-state
+  "Retrieves all virtual-machine records ."
+  []
+  (async/go
+    (let [
+          timestamp (str (t/now))                           ;; common timestamp for all snapshots
+          es-host (or (env/env :es-host) "127.0.0.1")
+          es-port (or (env/env :es-port-http) 9200)
```
loomis: Please use `:es-port` for this variable.  This is already what is used for the ssclj server.

global-state/jar/build.boot
```diff
@@ -0,0 +1,91 @@
+(def +version+ "3.37-SNAPSHOT")
+
+(set-env!
+  :project 'com.sixsq.slipstream/globalstate
```
loomis: I would propose using "metering" to replace "global state" everywhere.  This probably better describes what the service does.  This change should be carried throughout all of the files. (Apologies for the late opinion on the naming.)

global-state/jar/resources/log4j.properties
```diff
@@ -0,0 +1,12 @@
+# Root logger  set to INFO level of debugging.
+log4j.rootLogger=INFO, A1
+log4j.logger.com.sixsq.slipstream.globalstate=DEBUG
```
loomis: This should not be DEBUG by default.  Probably should be INFO.

global-state/jar/src/com/sixsq/slipstream/globalstate.clj
```diff
@@ -0,0 +1,83 @@
...
+          client (spandex/client {:hosts [(str "http://" es-host ":" es-port)]})
+          ch (spandex/scroll-chan client
+                                  {:url search-all-virtual-machines-url :body {:query {:match_all {}}}})]
+      (log/info "global state snapshots at " timestamp)
+      (loop []
+        (when-let [page (async/<! ch)]
+          (use-pagination client page timestamp)
+          (recur)))
+      )
```
loomis: Generally, don't leave hanging parentheses.

ssclj/jar/build.boot
```diff
@@ -156,7 +156,8 @@
                             #"config-hsqldb.edn"
                             #"log4j.properties"}
                  :invert true)
-           (aot :namespace #{'com.sixsq.slipstream.ssclj.app.main 'com.sixsq.slipstream.ssclj.usage.summarizer})
+           (aot :namespace #{'com.sixsq.slipstream.ssclj.app.main
+                             'com.sixsq.slipstream.ssclj.usage.summarizer})
```
loomis: Is the usage summarizer used in the metering service?  If so, why does it need to be AOT compiled?

global-state/jar/src/com/sixsq/slipstream/globalstate.clj
```diff
@@ -0,0 +1,83 @@
...
+    [clojure.tools.logging :as log]
+    [clj-time.core :as t]
+    [qbits.spandex :as spandex]
+    [clojure.core.async :as async]
+    [environ.core :as env]
+    [com.sixsq.slipstream.scheduler :as scheduler]))
+
+
+;;define the target index for the global state
+(def ^:const index-action {:index {:_index "resources-index" :_type "global-state-snapshot"}})
```
loomis: It probably would be better to parameterize the index to avoid having to recompile if we change it.

global-state/jar/src/com/sixsq/slipstream/globalstate.clj
```diff
@@ -0,0 +1,83 @@
...
+;;initial delay
+(def ^:const immediately 0)
+;;task interval
+(def ^:const every-minute (* 60 1000))
+;;global ES query to search every virtual-machine documents
+(def ^:const search-all-virtual-machines-url "resources-index/virtual-machine/_search")
```
loomis: Probably also better to parameterize this as well.  If we end up with storage or other resources to meter, do you think that we'll have a separate service for each or will extend this one to include those other resources?

global-state/jar/src/com/sixsq/slipstream/globalstate.clj
elegoff: OK, so I will add an environment variable like `METERING_ES_INDEX` defaulting to "resources-index"

global-state/jar/src/com/sixsq/slipstream/globalstate.clj
elegoff: My personal preference would be to extend this metering service so to include any other resources (like storage) in the future where this process would loop over a defined set of resources to meter (virtual machine, storage...). However my first attempt to do so, probably because of the current implementation of the scheduler

global-state/jar/src/com/sixsq/slipstream/globalstate.clj
elegoff: well noted, I wish cursive would so this automatically

global-state/jar/build.boot
elegoff: Done

ssclj/jar/build.boot
elegoff: Actually this file was accidentally touched, it should not be changed from its original content in master

global-state/jar/src/com/sixsq/slipstream/globalstate.clj
elegoff: As spandex relies on http protocol, I can source it as ES_PORT=9200, but ay any rate 9300 would not work

global-state/jar/resources/log4j.properties
elegoff: DONE

global-state/jar/src/com/sixsq/slipstream/globalstate.clj
elegoff: Created 2 env variables for parameterizing : 
METERING_INTERVAL=60000
METERING_SEARCH_URL=resources-index/virtual-machine/_search

Issue #1216, opened 2017-09-28
create quota resource


loomis:
Create an initial quota resource that can describe a quota at the levels of an organization and group.

PR #1220, opened 2017-09-28
quota resource implementation

loomis:
Connected to #1216.

### DIFF
--- a/ssclj/jar/test/com/sixsq/slipstream/ssclj/resources/quota_lifecycle_test.clj
+++ b/ssclj/jar/test/com/sixsq/slipstream/ssclj/resources/quota_lifecycle_test.clj
@@ -1,0 +1,142 @@
+(ns com.sixsq.slipstream.ssclj.resources.quota-lifecycle-test
+  (:require
+    [clojure.test :refer :all]
+    [peridot.core :refer :all]
+    [clojure.data.json :as json]
+    [com.sixsq.slipstream.ssclj.resources.quota :refer :all]
+    [com.sixsq.slipstream.ssclj.resources.lifecycle-test-utils :as ltu]
+    [com.sixsq.slipstream.ssclj.middleware.authn-info-header :refer [authn-info-header]]
+    [com.sixsq.slipstream.ssclj.app.routes :as routes]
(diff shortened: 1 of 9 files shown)


loomis:
Questions for reviewer:

 1. In the schema, the value for `collection` is technically the resource name.  That is, "VirtualMachine" rather than "virtualMachines".  We keep `collection` as the key or change to something else?
 2. The action/operation to evaluate the quota is called `evaluate`.  Is there a better verb for this?
 3. The `evaluate` action returns a copy of the quota resource with the keys `:current-all` and `:current-user` added.  These are the current usage for everything falling into the quota and the user's contributions to the quota (technically those visible to the user), respectively.  Are there better names for these?

konstan:
1. Here I guess we want to specify the type of the instances (`VirtualMachine`) in a named collection (`virtualMachines`).  Suggestions: `class`, `object-type`? I suspect there will be quotas for S3 etc.
2. IMHO, evaluation is a judgement, but here we are just collecting and returning the current aggregations as part of the quota.  I guess enforcement will happen later based on the collected values and the `:limit`.  So, I suggest `collect`.
3. LGTM.

ssclj/jar/test/com/sixsq/slipstream/ssclj/resources/quota_lifecycle_test.clj
```diff
@@ -0,0 +1,200 @@
...
+                            (ltu/is-status 200)
+                            :response)]
+
+          ;; check that the evaluate action returns the correct response
+          (is (= n-vm (:current-all eval-resp)))
+          (is (= (get freq "jane") (:current-user eval-resp)))
+          (is (= 100 (:limit eval-resp))))
+
+        ;; check that the evaluate action returns the correct response
+        (-> session-jane)
```
konstan: Request and assertions are missing?

konstan:
All the rest LGTM.

ssclj/jar/test/com/sixsq/slipstream/ssclj/resources/quota_lifecycle_test.clj
loomis: Cut and paste error.  The correct check of this is just above.

PR #1229, opened 2017-10-03
Resource tag should be plural

elegoff:


loomis:
@elegoff  To which resource does this apply?

elegoff:
"meterings"

Issue #1206, opened 2017-09-26
Add missing parts to the new VirtualMachine resource


schaubl:
- [ ] Service offer
- [ ] CPU, RAM, Disk, Instance type
- [ ] Price
- [ ] ...

PR #1235, opened 2017-10-05
Add missing parts to virtual machine

schaubl:
Connected to #1206

### DIFF
--- a/jar-persistence/src/main/java/com/sixsq/slipstream/persistence/VirtualMachine.java
+++ b/jar-persistence/src/main/java/com/sixsq/slipstream/persistence/VirtualMachine.java
@@ -49,18 +49,7 @@ public class VirtualMachine {
         public final String priceUnitCode;
 
         public ServiceOfferRef(String href) {
-            this.href = href;
-
-            this.resourceVcpu = null;
-            this.resourceRam = null;
-            this.resourceDisk = null;
-            this.resourceInstanceType = null;
-
-            this.priceCurrency = null;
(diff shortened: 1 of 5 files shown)


jar-persistence/src/main/java/com/sixsq/slipstream/persistence/VirtualMachine.java
```diff
@@ -20,8 +21,63 @@ public String toJson() {
...
+        @SerializedName("price:billingPeriodCode")
+        public final String priceBillingPeriodCode;
+
+        @SerializedName("price:freeUnits")
+        public final Float priceFreeUnits;
+
+        @SerializedName("price:unitCode")
+        public final String priceUnitCode;
+
         public ServiceOfferRef(String href) {
```
elegoff: Could be simpler with 
public ServiceOfferRef(String href) {
            this(href, null, null, null, null, null, null, null, null, null);
        }

jar-persistence/src/main/java/com/sixsq/slipstream/persistence/VirtualMachine.java
```diff
@@ -39,9 +95,16 @@ public boolean equals(CredentialRef credential) {
 
     public static class UserRef {
         public final String href;
+        public final String organization;
 
         public UserRef(String username){
             this.href = "user/" + username;
```
elegoff: //Same comment 
public UserRef(String username){
this(username, null);
}

Issue #1234, opened 2017-10-05
attach price to metering resource


elegoff:
Implementation of https://github.com/SixSq/tasklist/issues/1321

PR #1236, opened 2017-10-05
Issue #1234 attach price to metering

elegoff:
Connected to https://github.com/slipstream/SlipStreamServer/issues/1234


loomis:
I changed two things: 1) the code was throwing an exception if the billing period was something other than "MIN" or "HUR" and 2) the `:price` key is omitted if there is no value, rather than using an explicit null.

Issue #1249, opened 2017-10-16
allow test ssclj server to be used via java


loomis:
Create a static java class that will allow a test ssclj server to be started/stopped from the java side of things.

PR #1250, opened 2017-10-16
create jar for test ssclj server

loomis:
Connected to #1249.


konstan:
@loomis I just pushed 20c231d which is about generation of static methods on SscljTestServer and forceful refresh of ES indices.

Issue #1274, opened 2017-10-27
add vm mapping info in elasticsearch


loomis:
Add a documents in Elasticsearch with VM mapping information. This is to avoid a direct dependency between HSQLDB and the collector.

PR #1275, opened 2017-10-27
push vm mapping info to elasticsearch

loomis:
Connected to #1274.


loomis:
@0xbase12 Please take a look to see if the approach is correct, but do not merge.  This method is called from `com.sixsq.slipstream.run.RuntimeParameterMediator` and seems to have all of the necessary information.

0xbase12:
It's okey for me. Seems to do the trick

loomis:
@0xbase12 Should be OK now.  Take a look through and make sure everything looks OK. There are tests on the java and clojure sides.  The one thing not tested is whether the SscljProxy is working; hopefully that is well tested elsewhere.  

I've made it so that the VirtualMachineMapping resource can only be added, not edited.  We can discuss whether that is the correct behavior.

loomis:
@0xbase12 As discussed, the code now treats the runUuid and owner as mandatory parameters. It will also now set the serviceOffer field to null when the parameter value is either null or a blank string.

PR #1313, opened 2017-12-04
fix: bumped up version number of dummy connector java

konstan:


konstan:
This was already fixed directly in master.

PR #1315, opened 2017-12-05
Evaluate es test server

elegoff:
Connected to https://github.com/SixSq/tasklist/issues/1413


loomis:
My previous problems were because the branch hadn't been updated to the latest snapshot release.  I've merged the changes and refactored a bit the code.  Everything is working for me now.

However, there are two things I don't like about the PR. 

 1. The test dependencies, as they were, were leaking into the production code. I've marked them as test dependencies now, but then had to add them to the build file for the metering. I think that the test dependencies and probably also the spandex utils namespace should be extracted into a separate test artifact to avoid having fragile dependency configurations.

 2. The test server pollutes the output for the test cases with lots of logging, which, in most cases, will not be useful. Is there a way to remove this logging by default and only enable it if it is necessary?

PR #1358, opened 2018-01-18
refactor terminator

loomis:
Refactor the terminator to avoid loading all user objects for every garbage collection cycle.  Add some logging and rationalize others.


rpm/src/main/jetty-config/resources/logging.properties
```diff
@@ -17,14 +17,15 @@ java.util.logging.FileHandler.formatter=java.util.logging.SimpleFormatter
 
 # Mute org.restlet.ext.httpclient.internal.HttpMethodCall sendRequest
 java.util.logging.FileHandler.filter=com.sixsq.slipstream.util.RestletSendRequestMuteFilter
+java.util.logging.FileHandler.filter=
```
schaubl: What is the rationale behind this change ?

loomis: Good catch.  Cut & paste error.

Issue #1343, opened 2017-12-21
discuss whether to create a machine/device CIMI resource or to extend the connector one


cjdcordeiro:
this resource shall start including HW information, state, and possibly other (new) CIMI references to SLAs and owners.

When a NB is installed, a new CIMI resource UID is created. Then during the first NB boot, a callback (https://github.com/slipstream/SlipStreamServer/issues/1314) is made which will change this resource's state to "Active"

cjdcordeiro:
- [ ] use the connector resource

cjdcordeiro:
@loomis

PR #1361, opened 2018-01-19
extend nuvlabox cimi with activate/set-compromised operations

elegoff:
Connect to https://github.com/slipstream/SlipStreamServer/issues/1343


cimi/src/com/sixsq/slipstream/ssclj/resources/common/schema.clj
```diff
@@ -23,7 +23,7 @@
 ;; implementation-specific resource actions have a prefix
 (def ^:const impl-action-prefix (str slipstream-schema-uri "action/"))
 (def ^:const impl-prefixed-actions
-  #{:describe :validate :collect :execute})
+  #{:describe :validate :collect :execute :activate :set-compromised})
```
loomis: Perhaps "quarantine" as a verb instead.

Issue #1320, opened 2017-12-11
clean up korma/hsqldb dependencies


loomis:
Remove korma and hsqldb from namespaces that no longer use them. Move migration scripts to separate module to isolate those dependencies.

Issue #1276, opened 2017-10-30
split cimi-core and optional resources


loomis:
To allow for easier reuse of the core CIMI server, split the core CIMI feature from the resource definitions.

Consider renaming the core server from ssclj to cimi in this process.

loomis:
Given the changes to the user resource and the switch to leiningen for the build, this refactoring will have to redone based on the current code in master.

loomis:
@cjdcordeiro Please look at the current split of resources and provide feedback.

cjdcordeiro:
- what's https://github.com/slipstream/SlipStreamServer/tree/master/cimi-resources/test-resources for?
 - could the cloud-entry-point be moved out of the core? I know it needs to be there but it is just the naming, that might not fit all use cases
 - why do you need "job" in the core?
 - shouldn't we have all the specific session templates outside the core? (cyclone for example)
 - probably the same for credential-cloud...not sure about this one though...
 

Question: how do you configure which of the cimi-resources to include when launching the server?

loomis:
> what's https://github.com/slipstream/SlipStreamServer/tree/master/cimi-resources/test-resources for?

This directory contains test resources used when building/debugging the service during development.  These are "java" resources like logging configuration files, not CIMI resources.  That's an unfortunate clash in names. 

> could the cloud-entry-point be moved out of the core? I know it needs to be there but it is just the naming, that might not fit all use cases

Not if you want to maintain a working service.  This is actually the only resource that is required by the CIMI spec.  It would be possible to rename the URL for this resource, but I don't see that necessarily as being worth the additional trouble. 

> why do you need "job" in the core?

It is a standard CIMI resource, but optional. It is used to handle asynchronous tasks/actions. It can be moved. 

> shouldn't we have all the specific session templates outside the core? (cyclone for example)
> probably the same for credential-cloud...not sure about this one though...

Yes, we can move some subsets (or all) of the sessions and credentials.  Moving these requires a more detailed look at the code because there are shared utilities and dependencies (for example from the middleware) that may not be easy to disentangle without some refactoring.  In any case, I can start by moving the job resource.

> Question: how do you configure which of the cimi-resources to include when launching the server?

The service tries to load all the namespaces in the `com.sixsq.slipstream.ssclj.resources` package. Those namespaces just have to be accessible from the service's classpath. Essentially you drop a jar file with the resources in the "/lib" directory of the server.

cjdcordeiro:
[...]

PR #1365, opened 2018-01-23
separate migration scripts

loomis:
Connected to #1320 
Connected to #1276


loomis:
@konstan This shouldn't be merged until we define how to package/run the migration script.  I'd be interested in your opinions on the best way to achieve this, ideally without having to create an uberjar.

konstan:
For this migration script `com/sixsq/slipstream/ssclj/migrate/user_cred` it easy.  It is intended to be run from SS instance where all the dependencies are available on the host.  The migration procedure provides the details on how to build the class-path and launch the script.  See http://ssdocs.sixsq.com/en/latest/release_notes/candidate_releases.html#id9

In general, dependencies on the third-party libs can be added to a migration script's jar.  However, dependency on e.g. `com.sixsq.slipstream/SlipStreamCljResources-jar` should be marked as `provided`.  This will support the above intent if agreed.

The problems may come in the future where the host from which a script should be launched doesn't contain libs of all our components required for the migration.  E.g.: `CLASSPATH=/opt/slipstream/comp-x/lib/*:/opt/slipstream/comp-y/lib/*` with `comp-x` and `comp-y` at runtime being different micro-services running in different clusters.

At the same time - do we need to keep this script in the following releases?  It was only relevant for 3.38 to 3.39 migration and the RPM with the scrip is in the YUM repo anyways.  IMHO, it can be deleted or left in the repo, but skipped from building.  Anyways, it may not even compile against the next snapshot.

loomis:
@konstan Indeed perhaps the most pragmatic solution is to remove this migration script. However, we will still need a solution for migration of modules and deployments.  Maybe the easiest is to base the migration script on `boot` or `lein` and allow those tools to download the required dependencies. As long as we don't require enterprise packages in the migration scripts, the packages should be accessible.

PR #1393, opened 2018-02-13
use api key/secret for the deployments instead of machine cookie

konstan:
Connected to #1386


loomis:
Works for me now.  Can be merged when ready.

PR #1395, opened 2018-02-13
use User.roles for holding authz roles from the authn header

konstan:
Please don't merge it yet.  It still lacks full test with authn/z.

Connected to #1394

### DIFF
--- a/jar-persistence/src/main/java/com/sixsq/slipstream/persistence/User.java
+++ b/jar-persistence/src/main/java/com/sixsq/slipstream/persistence/User.java
@@ -542,7 +542,7 @@ public class User extends Metadata {
                     "query to find user cloud credentials.", e);
         }
         String resource = SscljProxy.BASE_RESOURCE + "credential?$filter=" + query;
-        Response resp = SscljProxy.get(resource, user.getName() + " USER", true);
+        Response resp = SscljProxy.get(resource, getCimiAuthnInfoUser(user), true);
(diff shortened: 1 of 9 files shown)


jar-persistence/src/main/java/com/sixsq/slipstream/persistence/User.java
```diff
@@ -949,18 +950,18 @@ private void checkNoForbiddenRoles(String roles) throws ValidationException {
         if (roles == null || roles.isEmpty()) {
             return;
         }
-        for (String role : roles.split(",")) {
+        for (String role : roles.split(" ")) {
             String trimedUppercaseRole = role.trim().toUpperCase();
             if (FORBIDDEN_ROLES.contains(trimedUppercaseRole)) {
```
loomis: This check will probably need to be removed as the ADMIN, USER, and ANON roles will probably already be present in the claims.

loomis:
The last two commits remove the calls to the user object for roles.

jar-persistence/src/main/java/com/sixsq/slipstream/persistence/User.java
konstan: Yes, I also noticed that and wanted to fix it when running the full test with authn/z.  I'm preparing my dev env for the test at the moment.

konstan:
@loomis This PR is ready for review.

Issue #1401, opened 2018-02-19
change default ACL on connector instance resource to allow VIEW for USER role


konstan:
```
{
        "principal": "USER",
        "right": "VIEW",
        "type": "ROLE"
      }
```
Related to https://github.com/slipstream/SlipStreamClient/issues/385

PR #1405, opened 2018-02-19
allow user to view connector instance configuration

konstan:
Connected to #1401

### DIFF
--- a/cimi-resources/src/com/sixsq/slipstream/ssclj/resources/service_attribute_namespace.clj
+++ b/cimi-resources/src/com/sixsq/slipstream/ssclj/resources/service_attribute_namespace.clj
@@ -89,7 +89,7 @@
 (defmethod crud/add resource-name
   [{{:keys [prefix uri]} :body :as request}]
   (if-let [id (colliding-id prefix uri)]
-    (response/response-conflict id)
+    (response/ex-conflict id)
     (add-impl request)))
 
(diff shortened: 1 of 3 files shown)


cimi-resources/src/com/sixsq/slipstream/ssclj/resources/service_attribute_namespace.clj
```diff
@@ -89,7 +89,7 @@
 (defmethod crud/add resource-name
   [{{:keys [prefix uri]} :body :as request}]
   (if-let [id (colliding-id prefix uri)]
-    (response/response-conflict id)
+    (response/ex-conflict id)
```
loomis: What's the reason for this change?  For me, this causes the lifecycle tests for the Service Attribute Namespace to fail.

konstan: It was failing for me with `response/response-conflict`, hence the change.  I might need to recompile the `response` dependency and retry.  I'll update you shortly.

konstan: screen shot 2018-02-20 at 08 22 44

konstan: Changed back to `response/response-conflict`.  All compiles and tests pass on my side.

Issue #1404, opened 2018-02-19
Garbage collector stacktrace in log in v3.45 and later


0xbase12:
From times to time, we have this stack trace inside error log of java server.

```
Exception in thread "async-dispatch-121" org.hibernate.LazyInitializationException: failed to lazily initialize a collection of role: com.sixsq.slipstream.persistence.Run.runtimeParameters, could not initialize proxy - no Session
	at org.hibernate.collection.internal.AbstractPersistentCollection.throwLazyInitializationException(AbstractPersistentCollection.java:575)
	at org.hibernate.collection.internal.AbstractPersistentCollection.withTemporarySessionIfNeeded(AbstractPersistentCollection.java:214)
	at org.hibernate.collection.internal.AbstractPersistentCollection.readElementByIndex(AbstractPersistentCollection.java:376)
	at org.hibernate.collection.internal.PersistentMap.get(PersistentMap.java:164)
	at com.sixsq.slipstream.statemachine.StateMachine.createNodeExtrinsicState(StateMachine.java:71)
	at com.sixsq.slipstream.statemachine.StateMachine.recreateStateFromRun(StateMachine.java:128)
	at com.sixsq.slipstream.statemachine.StateMachine.<init>(StateMachine.java:64)
	at com.sixsq.slipstream.statemachine.StateMachine.getStateMachine(StateMachine.java:120)
	at com.sixsq.slipstream.statemachine.StateMachine.createStateMachine(StateMachine.java:106)
	at com.sixsq.slipstream.util.Terminator.terminateInsideTransaction(Terminator.java:123)
	at com.sixsq.slipstream.util.Terminator.purgeRun(Terminator.java:109)
	at com.sixsq.slipstream.util.Terminator.purge(Terminator.java:74)
	at slipstream.async.garbage_collector$purge.invokeStatic(garbage_collector.clj:19)
	at slipstream.async.garbage_collector$purge.invoke(garbage_collector.clj:19)
	at slipstream.async.garbage_collector$collect_BANG_$fn__240$state_machine__7658__auto____243$fn__245.invoke(garbage_collector.clj:36)
	at slipstream.async.garbage_collector$collect_BANG_$fn__240$state_machine__7658__auto____243.invoke(garbage_collector.clj:36)
	at clojure.core.async.impl.ioc_macros$run_state_machine.invokeStatic(ioc_macros.clj:973)
[...]

PR #1406, opened 2018-02-20
log thread identities for garbage collection

loomis:
Connected to #1404.


loomis:
@0xbase12 The above changes are deployed on your SlipStream test instance.

PR #1412, opened 2018-02-26
finish containerization and tests with default es container

cjdcordeiro:
connected to https://github.com/SixSq/SlipStreamServer/issues/57


docker-compose.yml
```diff
@@ -0,0 +1,40 @@
+version: '3'
```
loomis: This file doesn't seem to be used within the build.  What is it used for?

cjdcordeiro: docker-compose.yml is the file used for deploying the CIMI server in containers. In can be seen as documentation

loomis: If it is documentation, then it is probably better in the ssdocs documentation (either the contributor's guide or administrator's guide). If there's a reason to keep it in the source code, then it should be moved to the cimi (cimi-resources, cimi-rpm, ?) submodule.

cjdcordeiro: maybe. But this is just for testing since the SS deployment doesn't rely on the Dockerized version of CIMI.

maybe put this into the README?

loomis: If this is intended purely for developers, then it can be added to a README for the CIMI server.

cjdcordeiro:
@loomis done

PR #1423, opened 2018-03-06
Docker docs

cjdcordeiro:


loomis:
@cjdcordeiro Following the procedure, I end up with everything running, but when trying `https://localhost/api/cloud-entry-point`, I get a "Bad Gateway" error.  Is there another endpoint, that I should be trying?

loomis:
Also trying `http://localhost:8201` returns an empty document.

cjdcordeiro:
https://localhost/api/cloud-entry-point should work. You should wait a bit. At least until you see something like this in the docker logs: `elasticsearch     | [2018-03-06T17:06:31,341][WARN ][o.e.d.i.m.StringFieldMapper$TypeParser] The [string] field is deprecated, please use [text] or [keyword] instead on [resourceURI]`

Issue #1422, opened 2018-03-05
make user ssh key not mandatory when creating user-param document 


konstan:

konstan:
The problem is on in the ssh key.  
Instead, after new user creation via OIDC, user-param resource cimi-side with the user General parameters is not created.  The General parameters that are seen on the old UI in the Profile are the default ones and available only in memory until they are edited and saved.

PR #1424, opened 2018-03-06
automatically add user exec params when new external user is created

konstan:
Connected to #1422


loomis:
Code builds. Demo worked.  Merging...

Issue #1442, opened 2018-03-14
allow reports bucket name to be configurable


konstan:
Implement via conf file to be able to start testing the service with functional tests in CI.

Later this should come from service configuration as per 
https://github.com/slipstream/SlipStreamServer/issues/1398 and
https://github.com/slipstream/SlipStreamServer/issues/1403.

PR #1443, opened 2018-03-14
allow to provide bucket name for storing reports in object store

konstan:
- obtain reports bucket name from configuration
- migrated config file from .ini to .edn
- package template for the config file with the cimi resources rpm

Connected to #1442


cimi-resources/src/com/sixsq/slipstream/ssclj/resources/external_object_report.clj
```diff
@@ -26,9 +26,8 @@
 ;;
 (def ^:const desc ExternalObjectReportDescription)
 
-
-;; S3 pre-signed-URLs
-(def ^:const report-bucket "slipstream-reports")            ;;single bucket containing reports, must exist
+;; single bucket containing reports, must exist
+(def reports-bucket (delay (:reportsBucketName (s3/object-store-config))))
```
loomis: Should this have a default value?

loomis:
@konstan Generally looks good and everything built for me.  There are a few suggested changes in 71d4fbe and a question on whether the bucket name should have a default value.

konstan:
@loomis I added the default for reports bucket name and accepting your changes.  (In fact I had the same `delay` based implementation at the beginning, but didn't like how the test looked :))
Anyways, everything builds for me.  I'm merging to see the result tomorrow.

PR #1448, opened 2018-03-15
fix inter container dependencies

cjdcordeiro:


loomis:
@cjdcordeiro I'm unable to test the deployment because of this error:
```
$ docker-compose up
Creating network "cimideployment_default" with the default driver
Pulling proxy (traefik:latest)...
latest: Pulling from library/traefik
ERROR: no matching manifest for linux/amd64 in the manifest list entries
```
Any ideas?

cimi-deployment/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+This document describes how you can run the CIMI server using Docker
+containers.
+
+## Prerequisites
+
+Ensure that [Docker](https://www.docker.com/) and [Docker
+Compose](https://docs.docker.com/compose/) installed.
+
+Make sure that you've allocated enough memory to Docker; 4-6 GB of
```
cjdcordeiro: this is just if you want to use ES, and want to give it 2GB of RAM. It is a very specific configuration. Should be more like a recommendation in case you want to use ES

cimi-deployment/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+You will find a `docker-compose.yml` file at the root of this
+directory.  This will manage the complete deployment including the
+CIMI servers and underlying database.
+
+If you wish, you can modify the contents of this file to customize the
+deployment.  For example,
+[Kibana](https://www.elastic.co/products/kibana) is optional.
+
+The version of the server is taken from the environment.  You must
```
cjdcordeiro: better to use a `.env` file, otherwise you'll need to export SLIPSTREAM_VERSION everytime you are in a new shell

loomis:
Works for me.

PR #1453, opened 2018-03-20
build arm image

cjdcordeiro:


cjdcordeiro:
connected to https://github.com/mF2C/cimi/issues/4

cjdcordeiro:
there's `buildArg` which might simplify this code duplication, but the image naming and base images would still be a problem

loomis:
@cjdcordeiro You should verify that the consolidated assembly descriptor works as expected for you.  It looked OK to me.

cjdcordeiro:
lgtm

Issue #1413, opened 2018-02-27
Ensure that s3 bucket is created


0xbase12:
Without the creation of report bucket:

Generation of presigned URL will not fail. But client will fail during creation of s3 object:

```curl https://slipstream-reports.sos-ch-dk-2.exo.io/test1\?X-Amz-Algorithm\=AWS4-HMAC-SHA256\&X-Amz-Date\=20180227T070956Z\&X-Amz-SignedHeaders\=host\&X-Amz-Expires\=-1519715396\&X-Amz-Credential\=EXO96d52217f24aafbd81f9a96a%2F20180227%2Fus-east-1%2Fs3%2Faws4_request\&X-Amz-Signature\=70b168fbf8b2936b9354b2a9e280a9ea8c9c3b9acbf076a4a05683de323bc369 --upload-file LICENSE.txt
<?xml version="1.0" encoding="UTF-8"?><Error><Code>NoSuchBucket</Code><Message>The specified bucket does not exist</Message><BucketName>slipstream-reports</BucketName><RequestId>c8b30139-8fbe-4ac8-bbe7-3ef4f5b9c0a8</RequestId><HostId>c8b30139-8fbe-4ac8-bbe7-3ef4f5b9c0a8</HostId></Error>%```

We have to manage report external object case and also others for new external object template

konstan:
Implemented here https://github.com/slipstream/SlipStreamServer/pull/1454/commits/872d15f6e21d43b40f998833d869dc16f79ebf3f

PR #1454, opened 2018-03-20
generic and reports specific external objects using cloud credentials

konstan:
The implementation of #1413 "Ensure that s3 bucket is created" will follow this one.

This can only be merged after connectors define `:objectStoreEndpoint` key: https://github.com/slipstream/SlipStreamConnectors/issues/213 and https://github.com/SixSq/SlipStreamConnectors/issues/155

Connected to #1443


cimi-resources/src/com/sixsq/slipstream/ssclj/resources/connector_template.clj
```diff
@@ -159,6 +166,15 @@
     :readOnly    false
     :order       10}
 
+   :objectStoreEndpoint
+   {:displayName "object.store.endpoint"
+    :type        "string"
+    :category    ""
+    :description "Cloud Object Store Service endpoint (e.g. http://s3.example.com:5000)"
+    :mandatory   true
```
loomis: Is the `:objectStoreEndpoint` key mandatory?  It is marked as mandatory here and is also mandatory in the "dummy" connector.  In contrast, it is optional in the "alpha" example connector template.

Issue #1480, opened 2018-04-16
[cimi-resources] using credentials for external object reports requires admin identity to fetch it


konstan:
For the moment, until uploading of user reports is using user credentials, expanding cloud credential in the case of external object report requires admin identity.

PR #1481, opened 2018-04-16
use admin identity to load creds in case of report external object

konstan:
Connected to #1480


konstan:
@loomis This was separately tested and the patch was applied on nuvla (where it works now with the intended not shared credential).

PR #1488, opened 2018-04-19
ignore special slipstream header on traefik

cjdcordeiro:


loomis:
Based on the documentation, this configuration appears to be correct.  I've not deployed it to test it.

Issue #1467, opened 2018-03-27
[cimi] [credential] Add an 'enabled' field in the Credential resource


schaubl:
This will allow to disable a credential. This will be useful, for example, to disable a credential if if doesn't work anymore (ex: the password is not valid anymore or cloud account deleted).

schaubl:
Doesn't apply anymore

PR #1518, opened 2018-05-15
Feature/credential resource with enabled flag

elegoff:
Connected to https://github.com/slipstream/SlipStreamServer/issues/1467


loomis:
Before merging this, I think that we need to have a wider discussion on the implications of having this flag.  First, is this really appropriate for all credentials?  (E.g. it isn't clear to me what a disabled public SSH key means.) Second, how does this affect the login process?  And third, can/should we ignore the enabled flag in certain circumstances?  (E.g. user requests the termination of a deployment but the associated credential is disabled.)

Issue #1598, opened 2018-07-10
termination problems because of credential access


loomis:
When a user has access to multiple cloud credentials for a given cloud, the lifecycle management code does not consistently use the correct ones.

In particular, the lifecycle management code may use one set of credentials for starting a deployment, but a different one for terminating it. On Exoscale for example, the terminate action will fail if those credentials are from different organizations.

loomis:
In addition, if a user authenticates via an external identity provider (particularly OIDC) and uses a shared credential, the terminate process fails.  Concretely, this is a problem with credentials that are shared with roles like "RHEA:can_deploy". On deployment, these roles are available and used to find the cloud credential.  These are not persisted in the user profile. On termination, the roles are taken from the user profile (which is usually empty). Consequently, the correct cloud credential cannot be found and the termination of the deployment fails.

loomis:
Proposed solution:

 - Add a run time parameter that gives the exact credential used to start a deployment.
 - On termination, identify the id of the credential saved in the deployment.
 - Retrieve (as super) and use that credential to terminate the deployment.

PR #1608, opened 2018-07-18
add/use correct credentials for terminate

loomis:
Connected to #1598


loomis:
@konstan Please review the changes before I attempt to test the changes.

jar-connector/src/main/java/com/sixsq/slipstream/connector/CliConnectorBase.java
```diff
@@ -416,6 +453,36 @@ static void genAndSetRunApiKey(Run run, User user, Map<String, String> environme
...
+		Set<String> cloudServiceNamesSet = new HashSet<String>();
+		for (String cloudServiceName : cloudServiceNames) {
+			cloudServiceNamesSet.add("connector/" + cloudServiceName);
+		}
+
+		// Get the credentials visible to the user
+		CloudCredentialCollection cloudCredentials = User.findCloudCredentials(user);
+
+		for (CloudCredential cloudCredential : cloudCredentials.getCredentials()) {
+			if (cloudServiceNamesSet.contains(cloudCredential.href)) {
```
konstan: It should be `cloudCredential.connector.href` (`href` is public on `connector`).  Or it can even be:
```
cloudServiceNames.contains(cloudCredential.connector.getRefResourceName())
```
i.e. w/o needing to build `cloudServiceNamesSet` out of resource IDs, because `getRefResourceName()` returns only the name part.  
See https://github.com/slipstream/SlipStreamServer/blob/26dfd688bc8f6113e979815f2c64cf750a7b0eaa/jar-persistence/src/main/java/com/sixsq/slipstream/credentials/HRef.java

loomis:
Suggested changes done.  Verified on a running test instance for both components and applications.

Issue #1611, opened 2018-07-20
add cimi deployment resources


loomis:
Implement an initial set of deployment resources that will allow the CIMI modules to be deployed. The design includes the following resources:

 - **DeploymentModelTemplate**: A template that takes a module reference and creates a new DeploymentModel. 
 - **DeploymentModel**: A "form" that contains the resolved information from a module and provides empty attributes (credentials, resource requirements, etc.) that can be filled in to then instantiate a running component/application.  The "start" action will instantiate a Deployment resource from the model.
 - **Deployment**: A running instance of a DeploymentModel. This document maintains the state information for the deployment. Note that individual run time parameters are handled as separate documents. Stopping a deployment may either be handled via an HTTP delete or a separate "stop" action. 
 - **DeploymentParameter**: The value of a single run time parameter.  This contains the deployment ID, node ID, parameter name, and parameter value, in addition to the common CIMI attributes.

schaubl:
Please find below some PR which might be useful to look at when doing the new module&deployment resources:
- Allow to deploy or build an application-component on multiple Clouds at the same time  (slipstream/SlipStream#128)
- Start a deployment with a credential instead of a connector (slipstream/SlipStreamServer#1622)
- SSCLJ security risk with slipstream-authn-info (SixSq/tasklist#1115)
- Allow to redefine parameters of the section "Cloud parameters" in the Run... dialog (slipstream/SlipStreamUI#358)
- Application workflow inheritance: multiple inheritance (slipstream/SlipStream#125)
- Allow to define a regex to validate Deployment parameters (slipstream/SlipStream#84)
- Allow to define the Type of each deployment parameters (slipstream/SlipStream#83)
- Authorization: separate the permission to BUILD and to RUN for Images (slipstream/SlipStream#82)
- Add the ability to use nodename parameter in the parameter mapping (slipstream/SlipStreamServer#14)

loomis:
New resources are in place.  Use of these resources through demonstrations will drive the further evolution.

PR #1612, opened 2018-07-20
deployment resource prototypes

loomis:
Connected to #1611


loomis:
Should not affect any other resources.  Merging to allow tests on Nuvla.

PR #1641, opened 2018-08-14
Issue #1666 configure metering to also snapshot records from bucky

elegoff:
Connected to https://github.com/SixSq/tasklist/issues/1666


loomis:
LGTM.

PR #1642, opened 2018-08-14
Add an optional unit attribute to the storage-bucket resource

elegoff:
Add optional unit in the bucky schema

Connected to https://github.com/SixSq/tasklist/issues/1666


cimi-resources/src/com/sixsq/slipstream/ssclj/resources/spec/storage_bucket.cljc
```diff
@@ -8,6 +8,7 @@
 (s/def ::credentials (s/coll-of ::cimi-common/resource-link))
 (s/def ::bucketName ::cimi-core/nonblank-string)
 (s/def ::usage pos-int?)
+(s/def ::unit ::cimi-core/nonblank-string)
```
cjdcordeiro: can you make this a selective list? only accepting certain values, like KB, MB, GB and B. this way we don't run into problems where someone added units in Kb/kb/kB...

Issue #1643, opened 2018-08-14
Storage bucket unit should be fixed to KiB


elegoff:
https://en.wikipedia.org/wiki/Kibibyte

PR #1644, opened 2018-08-14
Assume storage bucket unit is always KiB

elegoff:
Connected to https://github.com/slipstream/SlipStreamServer/issues/1643


cimi-resources/src/com/sixsq/slipstream/ssclj/resources/spec/storage_bucket.cljc
```diff
@@ -8,18 +8,18 @@
 (s/def ::credentials (s/coll-of ::cimi-common/resource-link))
 (s/def ::bucketName ::cimi-core/nonblank-string)
 (s/def ::usage pos-int?)
-(s/def ::unit #{"B","KB","MB", "GB"})
+(s/def ::unit #{"KiB"})
```
cjdcordeiro: it's probably better to change the name of the attribute `usage` to `usage_in_kibibytes`, and remove `unit`

loomis: That's OK for me as well, but the attribute names are camel cased, so "usageInKiB".

PR #1653, opened 2018-09-07
Feature/add identifier script

elegoff:
Connected to https://github.com/SixSq/tasklist/issues/1697

### DIFF
--- a/cimi-ident/README.md
+++ b/cimi-ident/README.md
@@ -1,4 +1,5 @@
-This script allows for a "manual" creation of a user-identifier document for 
+This script allows for a "manual" creation of a user-identifier document for
+
 - A user  
 - An instance (the OIDC organization)
   The external login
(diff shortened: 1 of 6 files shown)


loomis:
Having a utility for adding a user identifier is useful. However, I have a few concerns about this implementation:

 - It reuses the internal functions for creating a user identifier, but because of this, it needs a direct connection to the underlying Elasticsearch database. This means that it can only be used from machines that can connect directly to the Elasticsearch database.
 - By creating a standalone utility, the overhead in using the utility is increased, especially as it isn't included in any of the SlipStream packages. This means that it would probably only be used by the core developers that already have the entire development environment set up. 
 - As this can be accomplished via the CIMI API and much of the boilerplate is already provided by the SlipStream client commands, it would be better to include a new client command to link an identifier with a user account. This would make it generally available and reduce the maintenance overheads.

That being said, I think that it would be better to start by adding the relevant documentation to the administrators guide. This documentation could then describe how to link an authentication to a user account: 

 - Via the browser interface.
 - Directly with `curl`, and 
 - If necessary, through a new command in the client.

The documentation will also help define the workflow to automate the account linking process.

elegoff:
Closing and updating Admin guide to be done in https://github.com/slipstream/SlipStreamDocumentation/issues/145

Issue #1674, opened 2018-10-22
implement ResourceMetadata resource


loomis:
Implement the CIMI ResourceMetadata resource.

PR #1675, opened 2018-10-24
metadata resource

loomis:
Connected to #1674.


loomis:
Merging this initial implementation.  It will have to be updated once we understand how to generate the metadata information from the resource schemas.  A distinction will probably also need to be made between metadata for standard resources and metadata for new resources/templates created by the administrators.

Issue #1671, opened 2018-10-22
define metadata for cimi resource attributes


loomis:
Determine how metadata for CIMI resource attributes can be defined generally.

The current implementation provides a "describe" action for some resources that provides a general description of the parameters.  This is only implemented for a few resources, mainly commonly used templates.  

Aside from not being implemented globally, it doesn't provide a good description of the schema when there are nested elements (either lists or maps).  It also separates the description from the actual schema definition. 

Determine if something more like [spex](https://github.com/mpenet/spex) would provide a more general solution. This would have to work with both clojure and clojurescript.  If this approach is taken, then a mechanism for sharing this information between client and server must be implemented.  Perhaps either code sharing or use of the CIMI metadata mechanism.

In addition, the actual schema for the parameter metadata must be refined and extended. It must include ways to hint the type of input field, ordering, etc.  The same schema must be used for both the CIMI parameters and input/output parameters defined by users for components and applications.

loomis:
There appear to be three libraries that build on top of clojure spec that could be useful for maintaining metadata of the resource attributes:

 - [spec-tools](https://github.com/metosin/spec-tools)
 - [speculate](https://github.com/uswitch/speculate)
 - [spex](https://github.com/mpenet/spex)

Note that we already use spec-tools to generate the Elasticsearch mappings from the defined schemas. However, we need a mechanism of maintaining additional metadata for attributes that would help generate forms from the metadata.

loomis:
The spex library has a model that would be relatively easy to implement without changing major parts of the code.  It maintains a separate hierarchy for dependencies between the spec definitions and allows clojure metadata to be attached to the definitions.  Moreover, it provides mechanisms to recover metadata that is merged from all of the children of a given spec.

It is marked as "Subject to changes/breakage. Use at own risk.", but appears to have had a stable API since it first came out. Using this would allow the possibility of directly sharing the specs with associated metadata between the client and server.  ClojureScript support would have to be tested.  It also isn't clear that directly sharing the schemas would be the best option, especially as the resources may vary between SlipStream deployments. 

However, just testing the simple stuff:

```shell
user=> (-> (s/def ::zzz string?)
  #_=> (spex/vary-meta! {:something :you-need}))
:user/zzz
user=> (spex/meta ::zzz)
nil
```

doesn't appear to work.  If we decide to use this, this would have to be debugged.

loomis:
[...]

PR #1677, opened 2018-10-25
add metadata to specs

loomis:
Connected to #1671.


loomis:
One compromise was necessary in the schema definition when using spec-tools (and still having the tests and mapping generation work): the `properties` map now accepts only keyword keys.  Before it would accept either a keyword or a string.  In practice, this shouldn't have any practical effect on our usage.

Issue #1697, opened 2018-11-07
add acceptProtocols attribute to module


loomis:
An application may only be able to read data via specific protocols.  Add an "acceptProtocols" attribute that lists the protocols that can be used by the application. This is a list of opaque strings. The values used are by convention between the application developer and the data publisher.

Issue #1696, opened 2018-11-07
add acceptContentTypes attribute to module


loomis:
Add an attribute to the modules to specify the data format types that can be accepted by the application. The attribute value is a list of mime-types. Use the name `acceptContentTypes`.  The phrase "accept" is used as this is semantically similar to the HTTP "accept" header.

PR #1698, opened 2018-11-08
add fields to module for data format/protocol matching

loomis:
Connected to #1696 
Connected to #1697

### DIFF
--- a/cimi-resources/src/com/sixsq/slipstream/ssclj/resources/spec/module.cljc
+++ b/cimi-resources/src/com/sixsq/slipstream/ssclj/resources/spec/module.cljc
@@ -35,14 +35,14 @@
 ;; data management attributes
 ;;
 
-(s/def ::acceptContentTypes (s/nilable (s/coll-of ::cimi-core/mimetype :min-count 1 :kind vector?)))
+(s/def ::dataAcceptContentTypes (s/nilable (s/coll-of ::cimi-core/mimetype :min-count 1 :kind vector?)))
 
-(s/def ::acceptProtocols (s/nilable (s/coll-of ::cimi-core/token :min-count 1 :kind vector?)))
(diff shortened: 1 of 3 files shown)


loomis:
This PR adds two optional fields to the `module` resource: `acceptContentTypes` and `acceptProtocols`.  These  identify the data formats and data access protocols that are supported by the module.  

 - The implementation adds this to the core `module` resource.  Consequently, these fields are shared for all versions of the module.  This was done to facilitate the queries of the modules for specific data types/protocols.  Note that these fields are also allowed on image module, although they don't make much sense there. 
 - Modules that accept data must define both of these fields.  For modules that can analyze any data format, then the MIME type `application/octet-stream` should be used. A module must always list a concrete set of protocols.
 - The values for `acceptContentTypes` must be syntactically valid MIME type values.  They do not need to be officially registered values.  That is, values such as `application/x-my-special-type` can be used, although officially registered values are preferred if they are applicable.
 - The values for `acceptProtocols` are any valid token (sequence of characters without whitespace). The concrete values are agreed conventions between the producer and consumer of the data.  Values such as 'http+s3' or 'posix+nfs' are recommended.
 - With this schema, it is not possible to denote that a module requires multiple data formats simultaneously.

loomis:
@konstan Please review this PR along with the comments on the implementation.

loomis:
@konstan 

> Extra list dataRequireContentTypes or dataAcceptContentTypes as a map with :contentTypeName <"req" | "opt">?

How is that going to work in the deployment workflow?

loomis:
@konstan I've updated the attribute names.  Waiting on feedback for the required/optional before making any changes for that.

konstan:
@loomis After giving it a bit of thought - it's getting unnecessarily complex for the initial implementation. Lets not do this now.

Issue #1692, opened 2018-11-05
create new configuration template for nb identifiers


cjdcordeiro:
you can use the list of exoplanets from here:

```def get_exoplanet_hostname():
    """ Uses the public Exoplanet Archive's API to 
    get a random exoplanet hostname
    """

    # API call
    endpoint = 'https://exoplanetarchive.ipac.caltech.edu/cgi-bin/nstedAPI/nph-nstedAPI?table=exoplanets'
    raw_csv = StringIO(requests.get(endpoint).text)
    reader = csv.reader(raw_csv, delimiter=',')```

loomis:
Superseded by SixSq/SlipStreamConnectors#205.

PR #1708, opened 2018-11-09
Issue/#1692 add nuvlabox identifiers configuration

cjdcordeiro:
@schaubl because some of the other templates didn't have a lifecycle test, this one doesn't have it neither...

connected to #1692


cimi/src/com/sixsq/slipstream/ssclj/resources/spec/configuration_template_nuvlabox_identifier.cljc
```diff
@@ -0,0 +1,30 @@
+(ns com.sixsq.slipstream.ssclj.resources.spec.configuration-template-nuvlabox-identifier
+  (:require
+    [clojure.spec.alpha :as s]
+    [com.sixsq.slipstream.ssclj.resources.spec.configuration-template :as ps]
+    [com.sixsq.slipstream.ssclj.resources.spec.core :as cimi-core]
+    [com.sixsq.slipstream.ssclj.util.spec :as su]))
+
+(s/def ::name ::cimi-core/nonblank-string)
+(s/def ::identifiers (s/coll-of (s/keys :req-un [::name])))
```
loomis: Is there a reason to have a list of open maps?  If not, I'd suggest using this instead:
```
(s/def ::identifiers (s/nilable (s/coll-of ::cimi-core/nonblank-string :min-count 1 :kind vector?)))
```
if you want to keep the :identifiers field as required.  This necessitates having an explicit nil rather than an empty list. 

Better would be to make the field optional and then just use:
```
(s/def ::identifiers (s/coll-of ::cimi-core/nonblank-string :min-count 1 :kind vector?))
```
This avoids complications with XML representations of the document (even though we don't actually provide such representations at the moment).

cjdcordeiro: the reason for a list of maps is in fact (as suggested by @schaubl ) to keep the schema more flexible for future changes. We might want in the future to add something to that map, like "used" or "type"...etc.

as for making it optional, if it is optional, then what does the template do? that's the only attribute of the new template...all the others are common

loomis:
Replace by SixSq/SlipStreamConnectors#205.

Issue #1632, opened 2018-08-03
external object issues


loomis:
When trying to create/delete an external object resource using a credential for a cloud that doesn't define an objectStoreEndpoint, the server returns a 500 response. This may be the correct response, but in this case it should not create the external object resource.  It does and the new resource cannot be deleted. 

In addition, there appears to be an issue with storing external object contents to Exoscale. When trying with an image, the data was truncated.  Same was true with a small text file.

PR #1717, opened 2018-11-13
Issue #1632 external object issues

elegoff:
Connected to https://github.com/slipstream/SlipStreamServer/issues/1632

### DIFF
--- a/cimi-resources/src/com/sixsq/slipstream/ssclj/resources/external_object.clj
+++ b/cimi-resources/src/com/sixsq/slipstream/ssclj/resources/external_object.clj
@@ -384,7 +384,7 @@
 
 (defmethod delete-subtype :default
   [{:keys [objectType objectName bucketName objectStoreCred] :as resource} {{keep? :keep-s3-object} :body :as request}]
-  (when-not keep?
+  (when (and (not keep?) (s3/ok-to-delete-external-resource? resource request))
     (s3/delete-s3-object (s3/expand-obj-store-creds objectStoreCred request objectType) bucketName objectName))
   (delete-impl request))
 
--- a/cimi-resources/src/com/sixsq/slipstream/ssclj/resources/external_object.clj
+++ b/cimi-resources/src/com/sixsq/slipstream/ssclj/resources/external_object.clj
@@ -170,7 +170,7 @@
     (cond-> (dissoc resource :operations)
             (seq ops) (assoc :operations ops))))
 
-;; Just triggers the Session-level multimethod for adding operations
+;; Just triggers the resource-level multimethod for adding operations
 ;; to the Session resource.
 (defmethod crud/set-operations resource-uri
   [resource request]
--- a/cimi-resources/src/com/sixsq/slipstream/ssclj/resources/external_object.clj
+++ b/cimi-resources/src/com/sixsq/slipstream/ssclj/resources/external_object.clj
@@ -305,7 +305,7 @@
   (let [object-name (if (not-empty objectName)
                       objectName
                       (format "%s/%s" runUUID filename))
-        obj-store-conf (s3/expand-obj-store-creds objectStoreCred request objectType)]
+        obj-store-conf (s3/expand-obj-store-creds objectStoreCred)]
     (log/info "Requesting upload url:" object-name)
     (s3/generate-url obj-store-conf bucketName object-name :put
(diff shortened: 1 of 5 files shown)


cimi-resources/src/com/sixsq/slipstream/ssclj/resources/external_object/utils.clj
```diff
@@ -45,7 +51,109 @@
...
+      (check-authn-fn request)))
+
+
+(defn uploadable-bucket?
+  "When the bucket exists, but the user can't create the object.
+  The external object resource must not be created."
+  [obj-store-conf bucketName]
+  (let [upload-url (generate-url obj-store-conf bucketName "test-probe" :put {:ttl 3 :content-type "text/plain" :filename "test.txt"})]
+    (= 200 (http/put upload-url))))
+
```
loomis: I don't think that this implementation will work in practice.  The test file is never deleted. The second time the user tries to add an object to the same bucket, this request will fail and cause the function to incorrectly return false.  A cleaner solution with fewer error corner cases would be to check the ACL of the bucket itself, provided that the same calls work at both Amazon and Exoscale.

cimi-resources/src/com/sixsq/slipstream/ssclj/resources/external_object/utils.clj
```diff
@@ -45,7 +51,109 @@
...
+        s3client (get-s3-client obj-store-conf)]
+    (cond
+      ;; When the user requests to delete an object, but it no longer exists :
+      ;; The external object resource should be deleted normally.
+      (not (bucket-exists? s3client bucketName)) resource
+      ;; "When the user cannot access the bucket/object to delete it :
+      ;; the external object resource should not be deleted."
+      (not (authorized-bucket-operation? resource request a/can-modify?)) (logu/log-and-throw 403 (format "Deleting the bucket %s is not allowed" bucketName))
+      :all-ok resource)))
+
```
loomis: Shouldn't this also check the case where the bucket exists but the object does not?

cimi-resources/src/com/sixsq/slipstream/ssclj/resources/external_object.clj
```diff
@@ -409,9 +383,9 @@
           (fn [resource _] (:objectType resource)))
 
 (defmethod delete-subtype :default
```
loomis: I think that this implementation does not treat correctly the case where the S3 delete request throws an exception; at the very least an informative message should be added to the thrown exception before propagating this back up the stack. 

I also don't think that the logic is correct when the underlying bucket/object doesn't exist.  In this case, the flag will indicate that the external object can be deleted and then the code will actually try to delete the non-existent bucket/object, which will then throw an exception that causes the delete request to fail.

Also, I wonder whether there's any benefit is having this as a multi-method.  Neither of the existing implementations specialize the behavior and generally I wouldn't see where this is necessary.

cimi-resources/src/com/sixsq/slipstream/ssclj/resources/external_object/utils.clj
elegoff: https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-using-java-sdk.html

elegoff: Implementation changed in favor a of a HEAD request to the bucket. As stated by the Javadoc (https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/AmazonS3Client.html#headBucket-com.amazonaws.services.s3.model.HeadBucketRequest-) "This operation is useful to determine if a bucket exists and you have permission to access it."

cimi-resources/src/com/sixsq/slipstream/ssclj/resources/external_object.clj
elegoff: When bucket does exists, the cimi resource deletion now succeeds. A 403 is issued when the bucket ACL does not allow it.

cimi-resources/src/com/sixsq/slipstream/ssclj/resources/external_object/utils.clj
elegoff: done

Issue #1633, opened 2018-08-03
add public external objects


loomis:
Allow users to create *public* external objects that will have a permanent URL.  The workflow is the same as the standard one, except that the "ready" action will:

 - Set the ACL of the object to public read
 - Recover the permanent URL and store it in the external object resource

This can be used to maintain references to media files, documents, etc.  The permanent URL would allow them to be directly visible in browser interfaces.

loomis:
[S3 docs](https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-using-java-sdk.html) may be helpful.

PR #1735, opened 2018-12-07
Issue #1633 add public external objects

elegoff:
Connected to https://github.com/slipstream/SlipStreamServer/issues/1633


cimi-resources/src/com/sixsq/slipstream/ssclj/resources/external_object/utils.clj
```diff
@@ -234,3 +234,41 @@
...
+  "Returns the untouched resource. Side effect is only on S3 permissions"
+  [{:keys [objectStoreCred bucketName objectName] :as resource}]
+  (let [s3client (-> objectStoreCred
+                     (format-creds-for-s3-api)
+                     (get-s3-client))]
+    (try-set-public-read-object s3client bucketName objectName)
+    resource))
+
+
+(defn public-url
```
loomis: I don't think that this function has the behavior that you're expecting based on its use in the `add-public-url` function.

 - If the `.getURL` throws a runtime exception, then this will propagate up to the user.  Although, perhaps this will never throw.
 - If the URL is null, then the original `.toString` implementation will throw an NPE.  I've changed this to `str`, but in this case it will return an empty string for `(str nil)`, making the check in `add-public-url` useless.
 - It looks like `.getURL` will return a URL in all cases, whether the object is public or not.  Perhaps a more generic name for the `public-url` function would be more appropriate.

cimi-resources/test/com/sixsq/slipstream/ssclj/resources/external_object_public_lifecycle_test.clj
```diff
@@ -0,0 +1,133 @@
...
+                             (request abs-uri)
+                             (ltu/body->edn)
+                             (ltu/is-operation-present "ready")
+                             (ltu/is-status 200))
+
+            ready-url-action (str p/service-context (ltu/get-op uploading-eo "ready"))]
+
+
+        ;; Missing ACL should fail the action
+        (with-redefs [s3/set-acl-public-read throw-object-acl-does-not-exist]
```
loomis: Is this really the expected behavior from the AWS S3 client?  I would expect that the method would always return an ACL, whether the object is public or not.

Issue #1740, opened 2018-12-18
Add public external object URL for Slipstream logo in configuration


elegoff:
Current SS logo image is hardcoded and stored on server
It should be set a public external object URL accessible from configuration

PR #1742, opened 2018-12-18
Add a log configuration parameter to point to logo image

elegoff:
Connected to https://github.com/slipstream/SlipStreamServer/issues/1740


loomis:
After thinking about this in more detail, there's a general problem with using the configuration to identify the stylistic elements of the browser interface: normal and anonymous users normally do not have access to configuration resources.  As a result, their interface cannot discover the necessary information for changing the stylistic elements.

Probably, we'll need to have an implicit configuration where the browser interface will search for particular external object resources to use in the interface.  We'll need to discuss what conventions to use.

Issue #1751, opened 2019-01-18
remove describe action


loomis:
In preparation for using the resource metadata everywhere, remove the describe action from all resources and any supporting files.

loomis:
Unfortunately, the Java side of things still depends on the descriptions for serializing/deserializing the users, connectors, and configuration.  Consequently, these cannot be removed at this time.

PR #1752, opened 2019-01-18
remove describe from resources

loomis:
Connected to #1751


loomis:
Won't merge.  See parent ticket.

